PR #3, opened 2015-04-12
fix package.json warnings

psimyn:
empty devDependencies, add repository field - prevents warnings on npm install


choonkending:
@psimyn Thanks!

PR #6, opened 2015-04-27
Bump deps

psimyn:
Ignore font stuff, this should only change package.json

1. Remove sass-loader and node-sass from package json (currently breaking stuff)
2. npm outdated --parseable=true --depth=0 | awk -F'[:|@]' '{print $2}'
3. Update result


choonkending:
@psimyn Thanks! Was just about to bump mongoose up as you suggested, but you beat me to it!

PR #8, opened 2015-04-28
send serverside scss to node-noop

psimyn:
Fix from [here](http://jlongster.com/Backend-Apps-with-Webpack--Part-I#p26)


choonkending:
@psimyn Thanks!

PR #9, opened 2015-04-28
add eslint and jscsrc

psimyn:
change things just enough so everything can build


choonkending:
@psimyn Thanks! Tested and verified to be working. Will be fixing all those warnings in the next commit!

PR #12, opened 2015-06-02
Fixed typo in About page

kenfehling:


app/components/About.react.js
```diff
@@ -27,7 +27,7 @@ var About = React.createClass({
         </h1>
         <p className={classnames({
           move: this.state.opaque
-        })}>Ninja Ocean has is comprised of a team of passionate technology experts, aimed to do good.</p>
+        })}>Ninja Ocean is comprised of a team of passionate technology experts, aimed to do good.</p>
```
choonkending: @kenfehling Good find :+1:

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "react-webpack-node",
   "version": "1.1.0",
   "description": "A simple Node.js app using Express 4 with Webpack, React, alt, ImmutableJS",
-  "repository": "https://github.com/choonkending/react-webpack-node",
+  "repository": "https://github.com/kenfehling/react-webpack-node",
```
choonkending: Not sure about this one :)

package.json
```diff
@@ -37,11 +37,15 @@
...
+    "jquery": "^2.1.4",
+    "jquery-ui": "^1.10.5",
     "jscs-loader": "0.0.5",
     "jsx-loader": "^0.13.1",
     "keymirror": "^0.1.0",
-    "lodash": "^3.7.0",
+    "restify": "^3.0.3",
+    "restify-https": "0.0.3",
+    "fb": "git+https://github.com/Thuzi/facebook-node-sdk.git",
+    "lodash": "^3.9.3",
```
choonkending: Happy to bump the existing versions of jquery, lodash up, however I don't think we will be needing restify, restify-https, fb as of yet. :)

choonkending:
@kenfehling Good catch with the typo. Thanks for helping out!

With the fb login, I'd actually prefer to have it as part of `Login.react.js`. I like the idea of having another social login method, but I'd like the implementation to be more flux-like :) 

That is, onclick would trigger an action which we have defined under the actions/ folder, and that could make an API call. (I have not looked into FB social logins yet, but I imagine it would be that way). 

I've commented on the other parts of the code that might need to be changed as well. 

Perhaps you could create a PR for just the typo fix first, and then we can look into adding FB login separately as another branch? Let me know if you have any questions.

kenfehling:
Ohh... I didn't mean to include all that other stuff. I'm sorry, I don't know how Github works. I created the pull request for the typo and then kept playing with my branch.

choonkending:
No worries! Takes a while getting use to Github, but you'll be a master in no time!

Issue #20, opened 2015-06-07
Fixing images and resolving their paths


choonkending:
Currently the catch all in [routes.js](https://github.com/choonkending/react-webpack-node/blob/master/server/config/routes.js#L61) is not an elegant solution. Every request goes through the react-router unless specifically caught by the if statement.

Also, [initialImages.js](https://github.com/choonkending/react-webpack-node/blob/master/app/utils/initInitialImages.js) is definitely not a nice way to load images. We should be looking at [https://github.com/webpack/react-starter/tree/master/app/elements](https://github.com/webpack/react-starter/tree/master/app/elements) for some inspiration, or possibly html-loader.

choonkending:
Using html-loader is another issue altogether, but worth looking at nonetheless, as I think sokra has already addressed this issue in a cool way in his example [repo](https://github.com/webpack/react-starter)

choonkending:
This should be working now https://github.com/choonkending/react-webpack-node/commit/72fa56307ec9d19c82c3b24e51c9fcb278e6831e. 

@psimyn Can you try including images again? I remember you mentioning it was broken.

```javascript
import url from 'images/example.png';

•
•
•
render() {
   return (<img src={url} />);
}
```

PR #21, opened 2015-06-11
Universal JS

choonkending:
Fixes for the images problem in  https://github.com/choonkending/react-webpack-node/issues/20

Having favicon images copied over by webpack into the static/assets folder is not an elegant solution. Furthermore, initInitialImages.js feels like an unneeded file.

Outcome of discussions with @psimyn:
- [x] Change the term isomorphic to universal JS, because of [this](https://medium.com/@mjackson/universal-javascript-4761051b7ae9)
- [x] Remove use of swig rendering template engine and instead use html-loader as part of the webpack process. 
- [x] Remove initInitialImages.js. The problem here is that images are located in `app/images`, and we do not want to manually copy these images to the `static` folder which is served from express. Webpack url-loader and file-loader can solve this, i.e. having `<link rel>`elements reference the correct image path determined by webpack. 
   - So we considered adding another React component for favicons and editing the HTML within `utils/UniversalRenderer`. The problem from this approach is that it makes `utils/UniversalRenderer.js` less atomic in its function. Ideally UniversalRenderer should be supplied an App, and it renders the App depending on whether it s on the client or server. 
   - We are now considering adding rendered html for favicons into the HTML when express serves the html response.


app/elements/Favicon.react.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+import url from 'file!images/chrome-ninja192-precomposed.png';
+
+class Favicon extends React.Component {
```
psimyn: why are there tab characters here?

psimyn:
are the other images going to be added as components, or favicon work for the os-specific guff as well?

app/elements/Favicon.react.js
choonkending: Good catch. What a derp.

choonkending:
@psimyn My plan involves creating 3 files:
1. A configuration file, something like this:
```
{
  apple: 'apple.png',
  chrome: 'chrome.png'
}
```
2. A FaviconGenerator in utils, which will:
    -  Iterate through the config file and pass in the url props into Favicon.react.js
    - Render the output of that into a string and return it

3. Make Favicon.react.js import it's props value and then render the link based on that.

choonkending:
Update on my local branch:

faviconfig.js
```javascript
const config = [
  {
    type: "link",
    rel: "icon",
    sizes: "192x192",
    filename: "chrome-ninja192-precomposed.png"
  },
  {
    type: "link",
    rel: "apple-touch-icon",
    filename: "apple-ninja152-precomposed.png"
  },
  {
    type: "meta",
    name: "msapplication-TileImage",
    filename: "ms-ninja144-precomposed.png"
  }
];

export default config;
```

FaviconGenerator.js
```javascript
import React from 'react';

import Favicon from 'elements/Favicon.react';
import config from 'faviconfig.js';

const imagePath = 'file!images/';

// TODO: Add polyfill for Array.prototype.map()
const renderedFavicons = config.map((fav) => {
  let fullImagePath = imagePath + fav.filename;
  import url from fullImagePath;
  return React.renderToString(<Favicon fav={fav} url={url}/>);
}).join('');

export default renderedFavicons;

```

Problem with this approach: [This guy](http://www.2ality.com/2014/09/es6-modules-final.html) shows that we can `import url from <variable>` unless we use `System.import().then()` which is a API that returns a Promise. Webpack in turn might not work well with this, stated here https://github.com/webpack/webpack/issues/816

choonkending:
We can use `require.ensure` which should work.

choonkending:
The irony: https://github.com/nfl/react-helmet

PR #22, opened 2015-06-15
Webpack resolve

psimyn:
Resolving paths in the app so that we remove pesky `..` and `./` import statements.


choonkending:
@psimyn Do you drive a sports car? Cause that was too fast and too furious! :racehorse:

choonkending:
@psimyn Good work. Merging.

PR #34, opened 2015-07-09
initial webpack-dev-server setup

joshhornby:
Steps needed to run:

```npm run build```
```npm run devserver```

As you can see this will point to the public folder but not sure how we should set it up so it can point to ```server/index.js```

Issue #33, opened 2015-07-08
Webpack dev server


joshhornby:
Just wondering if anyone has successfully managed to set up a [webpack-dev-server] (http://webpack.github.io/docs/webpack-dev-server.html) for this project?

choonkending:
@joshhornby Thanks for asking! We haven't set it up yet but it's on our to-do list. Was hoping for dev server and React hot loader to speed up dev work. 

Would you be interested in working on this? I'll bump this up in priority. :)

joshhornby:
I was hacking around earlier and could get webpack-dev-server to work with hot loading but my issue was I didn't know how to run the ```server/index.js``` in webpack. Any ideas @choonkending

choonkending:
None yet, (was reading more about webpack-dev-server on the tram to work) :) If you have a branch up I can have a look!

joshhornby:
I have pushed up what I have. https://github.com/choonkending/react-webpack-node/pull/34 We just need to get the node app into the webpack config and it should work.

PR #37, opened 2015-07-11
React Hot Loader

choonkending:
Referring to https://github.com/choonkending/react-webpack-node/issues/33 and PR here https://github.com/choonkending/react-webpack-node/pull/34

Changes 
- [x] Moved webpack.config.js to webpack folder
- [x] Have a dev-server for hot reloading
- [x] Update package.json
- [x] Update README

To run:
1. `npm install` the additional packages required
2. `npm run buildDev` for a different webpack build
2. `npm run dev` to start the server. Make any changes to your Components and watch them change live!

Note:
1. Does not work with css changes (yet). That can potentially be resolved after we start using css modules
2. If you are making changes to store, I am not sure how this would work out yet. Might have to make our altStores hotStores described https://github.com/goatslacker/alt/issues/85

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,15 +5,20 @@
   "repository": "https://github.com/choonkending/react-webpack-node",
   "main": "index.js",
   "scripts": {
-    "start": "node server/index.js",
-    "build": "webpack --progress",
-    "watch": "webpack --watch --progress",
+    "start": "NODE_ENV=production node server/index.js",
+    "dev": "NODE_ENV=development node server/index.js",
+    "build": "webpack --progress --colors --config ./webpack/webpack.config.js",
+    "buildDev": "webpack --progress --colors --config ./webpack/webpack-hot-dev-server.config.js",
+    "watch": "webpack --watch --progress --config ./webpack/webpack.config.js",
     "test": "echo \"Error: no test specified\" && exit 1",
-    "postinstall": "webpack -p"
+    "postinstall": "webpack -p --config ./webpack/webpack.config.js"
   },
   "author": "Choon Ken Ding",
   "license": "MIT",
-  "devDependencies": {},
+  "devDependencies": {
+    "react-hot-loader": "^1.2.8",
+    "webpack-dev-server": "^1.10.1"
+  },
   "engines": {
     "node": "0.10.x"
   },
@@ -38,6 +43,7 @@
     "extract-text-webpack-plugin": "^0.8.1",
     "file-loader": "^0.8.4",
     "html-loader": "^0.3.0",
+    "http-proxy": "^1.11.1",
     "immutable": "^3.7.4",
     "iso": "^4.1.0",
     "jquery": "^2.1.4",
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "main": "index.js",
   "scripts": {
     "start": "NODE_ENV=production node server/index.js",
-    "dev": "NODE_ENV=development node server/index.js",
+    "dev": "NODE_ENV=development npm run buildDev && node server/index.js",
     "build": "webpack --progress --colors --config ./webpack/webpack.config.js",
     "buildDev": "webpack --progress --colors --config ./webpack/webpack-hot-dev-server.config.js",
     "watch": "webpack --watch --progress --config ./webpack/webpack.config.js",
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
     "build": "webpack --progress --colors --config ./webpack/webpack.config.js",
     "buildDev": "webpack --progress --colors --config ./webpack/webpack-hot-dev-server.config.js",
     "watch": "webpack --watch --progress --config ./webpack/webpack.config.js",
+    "watchDev": "webpack --watch --progress --config ./webpack/webpack-hot-dev-server.config.js",
     "test": "echo \"Error: no test specified\" && exit 1",
     "postinstall": "webpack -p --config ./webpack/webpack.config.js"
   },
--- a/Changelog.md
+++ b/Changelog.md
@@ -36,3 +36,9 @@
 - Renamed IsomorphicRouterRenderer to UniversalRenderer
 - Removing views folder from server as we no longer need them
 - Removing socket.io dependency as the examples no longer supports it
+
+1.1.3
+=======
+
+- Added [react-hot-loader](https://github.com/gaearon/react-hot-loader) for development speed
+- Moved webpack config files into the /webpack folder
--- a/README.md
+++ b/README.md
@@ -47,8 +47,7 @@ I am working to document this repo extensively so it would be easy for both begi
 
 We use [react-hot-loader](https://github.com/gaearon/react-hot-loader), which is about the greatest thing that has ever happened. No browser refreshes needed.
 
-1. `npm run watchDev` to watch and recompile for changes using the `webpack-hot-dev-server.config.js` file.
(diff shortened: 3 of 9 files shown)


choonkending:
@psimyn Please have a look and test it out!

package.json
```diff
@@ -5,15 +5,20 @@
   "repository": "https://github.com/choonkending/react-webpack-node",
   "main": "index.js",
   "scripts": {
-    "start": "node server/index.js",
-    "build": "webpack --progress",
-    "watch": "webpack --watch --progress",
+    "start": "NODE_ENV=production node server/index.js",
+    "dev": "NODE_ENV=development node server/index.js",
+    "build": "webpack --progress --colors --config ./webpack/webpack.config.js",
+    "buildDev": "webpack --progress --colors --config ./webpack/webpack-hot-dev-server.config.js",
```
joshhornby: Do you need to add a ```--watch``` here? its the only way I can get it to work correctly. Other wise some of the data isn't getting reloaded - for example props on a component.

joshhornby:
Also it seems that the node application doesn't refresh on file save?

choonkending:
@joshhornby It works for me on my end (refreshing on save). 

I ran `npm run buildDev` once , and then ran `npm run dev` to start the server.

Haven't had a chance to look into this since the weekend. Can you list down the steps of what commands you used? I'll have a look when I get the chance. :)

package.json
choonkending: `npm run dev` should recompile and change your page on file save, without a `--watch`.

joshhornby:
Sure thing:

npm run buildDev
npm run dev

I have tried ```npm run buildDev``` with and without the watch flag. The react app updates correctly but not the node side.

choonkending:
Yup I got the same problem. It does not update reliably and the server rendered template does not reflect the changes. 

Sorry bout the late reply. A bit caught up with other projects.

choonkending:
@joshhornby I had a look into this. The approach is to use `dev-server.js` as a proxy for our static assets. webpack-dev-server does not write to disk, read the issue created [here in webpack-dev-server's repo](https://github.com/webpack/webpack-dev-server/issues/62). 

> The webpack-dev-server doesn't write files to disk. You can use webpack --watch which writes files to disk. But why would you need to access the generated assets? This shouldn't be required.
When using with an existing server you should thread the webpack assets as blackbox that exists at a special URL.

But since we also have React rendering on the server side, we need to be able to reload and see the changes we made. So you can `npm run watchDev` to work with reloads (although you will see a checksum error appearing in the console. This is because although the server rendered template is served from our backend server, but the `app.js` being served is still the version being served by our dev-server. This is nothing to worry about because it will not happen in production mode).

How about styles in dev? Refer to [this issue](https://github.com/webpack/extract-text-webpack-plugin/issues/30)

> You shoudn't use the extract-text-webpack-plugin in development. 
 
So in dev mode, I've removed `ExtractTextPlugin`. There will be a FOUC(Flash of unstyled content) but you will be able to make changes to scss files and get a neat hot reload.

@psimyn Acceptable?

So now the workflow would be:
1. `npm run watchDev`
2. `npm run dev`

joshhornby:
Will this solve the issue of node not reloading on file save? I am currently doing 

```"webpack --watch --progress --colors --config ./webpack/webpack-hot-dev-server.config.js```

and in another tab

```NODE_ENV=development node server/index.js```

**Comment by choonkending** [comment]
Yes it should. Running the `watch` command will save the changes to file. 

To clarify, hot loader doesn't save changes to file, it replaces the module you are currently working on to speed up your dev work. :)

**Comment by choonkending** [comment]
Will be merging at the end of today :)

**Comment by joshhornby** [comment]
I can confirm that the node application doesn't automatically reload. No sure if this is the same for you?

**Comment by choonkending** [comment]
@joshhornby Do you mean that the server does not reload? (Does the browser reload?) If you could tell me what you're running and expecting to see it will help me debug the problem.

**Comment by joshhornby** [comment]
The server and browser both don't reload. The only way to get the node app to reload is to kill the dev ```npm run dev``` command.

I am running

```npm run watchDev```
```npm run dev```

and I have also tried the ```npm run watch``` command.

**Comment by choonkending** [comment]
Hmm. I'll double check again, and get @psimyn to test it. 

Also, do you have any error messages/console logs in the browser?

Can you try `npm install` again, and run `npm run watchDev` and `npm run dev` again?

Thanks! Haha what a tough nut to crack

**Comment by choonkending** [comment]
I've removed some of the unneeded options on my `dev-server`.

Can you pull the latest changes and test this out? :racehorse: 

First off just try running `npm run dev`. The components should be hot reloaded. I git cloned this to another folder and `npm install`ed this again to test this lol.

If you want to reload and see the changes, then run `npm run watchDev`.

**Comment by joshhornby** [comment]
I pulled in the changes, but when I just ran ```npm run dev``` which was an alias of ```NODE_ENV=development npm run buildDev && node server/index.js``` the server didn't boot up so I could get to my localhost:3000. No issues in the terminal though.

So instead I reverted to the old commands ```npm run watchDev npm run dev``` and see this error in the terminal ```GET http://0.0.0.0:8080/socket.io/?EIO=3&transport=polling&t=1437569304215-30 net::ERR_CONNECTION_REFUSED```

**Comment by choonkending** [comment]
The server should start running on `npm run dev`, it might take a while because it's bundling, and you should be able to access localhost:3000.

Also, the dev-server that proxies our requests is on port 3001. My network tab shows `Request URL:http://localhost:3001/socket.io/?EIO=3&transport=polling&t=1437570824401-2&sid=eF0m7YruLHbWPhpzAAAB`.

My first commit was on port `8080`, but in subsequent commits I've switched to `3001`. If you're running on this repository, can you run the following commands to ensure you've pulled the latest? I got a feeling you're running on the initial changes to this pull request.

`git pull origin feature/react-hot-loader`. Otherwise, make sure `express.js` matches
```javascript
    app.all('/assets/*', function(req, res) {
      proxy.web(req, res, {
          target: 'http://localhost:3001'
      });
    });
```

and webpack-hot-dev-server.config.js has the entry point:

```javascript
var WEBPACK_HOST = "localhost";
var WEBPACK_PORT = 3001;

•
•
•

entry: {
      app:[ 'webpack-dev-server/client?http://' + WEBPACK_HOST + ":" + WEBPACK_PORT,
       'webpack/hot/only-dev-server',
        "./app" ]
    },
```

joshhornby:
Ah yes, I have made sure the repo I am running on is running with the latest code. 

But again I can't just run ```npm run dev``` and have the site boot up. And unfortunately this still doesn't work node side, React works correctly but if I place a console log inside a node file, the only way to see the change is to kill the server and then boot up again/

choonkending:
I see. If you run `npm run dev` and go to localhost:3000 do you see
anything? Just clarifying cause the hot loader does not open a browser tab
for you.

Also any changes to the nodejs  file (not a file that is bundled by
webpack) shouldn't trigger a reload or restart of the server, as that s not
what webpack dev server and hot reloader does. Webpack dev server + hot
loader work helps with live editing and not having to reload your browser
(not the server).

If you want nodejs server to reload on file save (placing a console log
inside the server file), we could use `pm2` or `nodemon` for that I
believe.

psimyn:
you can watch the server folder for changes by updating dev script to be something like
```
nodemon --watch server server/index.js
```

but restarting the server will break live reloading - because dev server doesn't write to disk even refreshing the page won't update changes.

Not sure if there is a server hot reloader.

joshhornby:
In that case the hot loader is working correctly. I will use nodemon :)

choonkending:
Cool! :) Merging! :shipit:

Issue #51, opened 2015-08-10
Error: Unhandled aborted transition! Reason: [object Object]


joshhornby:
As mentioned here https://github.com/choonkending/react-webpack-node/issues/50#issuecomment-129511751 I am having issues when trying to set up an authenticated route.

You can see [in the following code example](https://github.com/joshhornby/react-webpack-node/commit/a7701eaa7004c069c124ffe766da03baed7d942f) - if you try and visit the about route via the nav bar you will successfully see the login page such as `http://localhost:3000/login?nextPath=%2Fabout`

But if you go to `http://localhost:3000/about` directly the following error is shown:

```
Error: Unhandled aborted transition! Reason: [object Object]
   at Function.React.createClass.statics.handleAbort (/Users/joshuahornby/code/react-webpack-node/node_modules/react-router/lib/createRouter.js:262:55)
   at Function.dispatchHandler (/Users/joshuahornby/code/react-webpack-node/node_modules/react-router/lib/createRouter.js:379:20)
   at /Users/joshuahornby/code/react-webpack-node/node_modules/react-router/lib/createRouter.js:349:29
   at /Users/joshuahornby/code/react-webpack-node/node_modules/react-router/lib/Transition.js:64:41
   at /Users/joshuahornby/code/react-webpack-node/node_modules/react-router/lib/Transition.js:69:9
   at Function.Transition.to (/Users/joshuahornby/code/react-webpack-node/node_modules/react-router/lib/Transition.js:72:15)
   at /Users/joshuahornby/code/react-webpack-node/node_modules/react-router/lib/createRouter.js:348:22
   at Function.Transition.from (/Users/joshuahornby/code/react-webpack-node/node_modules/react-router/lib/Transition.js:51:15)
   at Function.dispatch (/Users/joshuahornby/code/react-webpack-node/node_modules/react-router/lib/createRouter.js:345:20)
   at Function.refresh (/Users/joshuahornby/code/react-webpack-node/node_modules/react-router/lib/createRouter.js:396:16)
   at Function.run (/Users/joshuahornby/code/react-webpack-node/node_modules/react-router/lib/createRouter.js:392:16)
   at Object.runRouter [as run] (/Users/joshuahornby/code/react-webpack-node/node_modules/react-router/lib/runRouter.js:45:10)
   at renderToMarkup (/Users/joshuahornby/code/react-webpack-node/public/assets/app.server.js:132:29)
[...]

Issue #50, opened 2015-08-10
Hitting refresh on nested routes


kkotwal94:
I have my routes set up in a way where nested routes exist. When hitting refresh on non nested routes it loads the stores and html fine with this: 

```
app.get('*', function (req, res, next) {
    
    var html = App(JSON.stringify(res.locals.data || {}), req.url);
    html = html.replace("TITLE", Header.title)
                .replace("META", Header.meta)
                .replace("LINK", Header.link);

    res.contentType = "text/html; charset=utf8";
    res.end(html);
  });
```
So I decided to mess with a bit and try to understand why none of the css or data was loading on the refresh of nested route that is dynamic (based off '/x/:id') so I decided to do something like this. 
```
app.get('*', function (req, res) { 

    Router.run(routes, req.path, function (Handler, state) {
        var element = React.createElement(Handler);
        var html = React.renderToString(element);
        res.render('main', { content: html });
    });
});
```
<strong>I just wanted to know if what was written in the boilerplate was supposed to deal with nested routes, if so I'll know my problem lies elsewhere. I haven't tested this yet, because I have to require my 'route' here on the server as well. I also have little idea of what is going on in the boilerplay app.get(*)</strong>

joshhornby:
On a separate but somewhat related note I am having issues with the render as well - I am trying to create a protected route following something like this http://www.aaron-powell.com/posts/2015-06-08-implementing-security-in-react-with-react-router.html

But my ```willTransitionTo``` method is failing with the following error ```Error: Unhandled aborted transition! Reason: [object Object]``` 

I have found a package that will solve this error https://github.com/ericclemmons/react-router-server-location but it will mean using a more decoupled router like the one above. I can create a separate issue for this is you like?

choonkending:
@kkotwal94 the boilerplate was meant to deal with nested routes initially, but I'm not sure it applies in your case. I'll have a look after I finish the protected routes feature #47 and get back to you on that :)

kkotwal94:
Sure I'm going to continue to debug and figure it out, and thanks.

kkotwal94:
```
app.get('*', function (req, res) { 

    Router.run(routes, req.path, function (Handler, state) {
        var element = React.createElement(Handler);
        var html = React.renderToString(element);
        res.render('main', { content: html });
    });
});
```
[...]

Issue #47, opened 2015-08-06
React Router with protected routes example


choonkending:
This has been a frequent question. Gotta get this one out!

joshhornby:
This would be great, perfect timing as I need to get this sorted in an application of mine!

choonkending:
Reopen if we find any issues. :)

PR #52, opened 2015-08-11
Feature/protected routes

choonkending:
# Related Issues
#47 #50 #51

Protected routes example. I'm still thinking what's the best way to render user specific data.

Also note: `willTransitionTo` needs to be changed once react-router 1.x is stable.


choonkending:
Thinking of a better solution: which is to grab the state from store, but `willTransitionTo` won't work in this case. 

https://github.com/rackt/react-router/pull/590#issuecomment-111910388 seems like a very convincing argument to update bump react-router to `1.x`.

joshhornby:
I don't see a reason not to bump up to `1.x` - Seems a sensible approach.

I think you mentioned it above but once I redirect a user to the page they initially requested I don't have the full user object just the state which has been set by the login. But I think you know about this!

choonkending:
@joshhornby Yep. I'll be working on this tonight and hopefully can get it done for you to test.

choonkending:
Might still get an error saying: 

```Failed propType: Invalid prop `location` supplied to `Router`, expected instance of `Location```

But it won't cause any problems (that I can find so far)
https://github.com/rackt/react-router/issues/1524

**Comment by choonkending** [comment]
One last checkbox:
- [ ] Add the auth to UserStore object, so we can grab it from client and server. Then this PR will be complete. 

Yay. Sleep time.

**Comment by choonkending** [comment]
@joshhornby I've done it. You can have a play with it now. 

If you want to access the UserStore object, your Dashboard component needs to listen to the Store. (In the future I'll be using AltContainer for the entire App).

File: app/utils/UserWebAPIUtils.js [hunk 1]
```diff
@@ -35,6 +35,13 @@ const utils = {
       contentType: 'application/json',
       data: JSON.stringify(data)
     });
+  },
```
joshhornby: This can be deleted.

server/config/routes.js
```diff
@@ -15,6 +15,7 @@ module.exports = function(app, passport) {
   app.post('/login', users.postLogin);
   app.post('/signup', users.postSignUp);
   app.get('/logout', users.getLogout);
+  app.get('/loggedin', users.getLoggedIn);
```
joshhornby: This can be deleted.

server/controllers/users.js
```diff
@@ -56,3 +56,16 @@ exports.postSignUp = function(req, res, next) {
   });
 };
 
+/**
```
joshhornby: This can be deleted.

joshhornby:
@choonkending Looks really good and seems to work for me.

Might this be a good chance to also solve this issue https://github.com/choonkending/react-webpack-node/issues/56?

Should be a case of creating another ```<Route>``` in the routes file such as 

```
<Route name="unauthorisedApp" path="/auth" component={UnauthorisedApp}>
   <Route path="logout" name="logout" component={Logout} />
   <Route path="login" name="login" component={Login} />
</Route>
```

The only issue I can't work out is the correct place to 'redirect' the user to their route. Good work though :+1:

choonkending:
@joshhornby Thanks for that!

Hmm. I think we should probably address #56 as another PR, so this won't be a never-ending PR.

I agree with the login component, I think its current position isn't ideal.

> The only issue I can't work out is the correct place to 'redirect' the user to their route.

What do you mean? As in a Redirect route to when they hit a 404? Or when they meet a certain condition?

joshhornby:
@choonkending Fair point and I agree.

Ignore me for now, this is an issue I am having with my set up!

choonkending:
Alright. I will merge this into master in a couple of hours. :)

This was a good journey with react-router beta. :football:

Issue #58, opened 2015-08-20
requireAuth not getting hit on page load


joshhornby:
The `requireAuth` function (https://github.com/choonkending/react-webpack-node/pull/52/files#diff-70012bb93dfce2be702a0251971c1db5R13) which was introduced in the PR https://github.com/choonkending/react-webpack-node/pull/52 works great if I click on a route with has a `onEnter` attached to it. 

But if I visit a link via the url for example `localhost:3000/dashboard` I am redirected but the following line https://github.com/choonkending/react-webpack-node/pull/52/files#diff-70012bb93dfce2be702a0251971c1db5R15 is never hit. This means the app doesn't know where to redirect to once the user has logged it.

choonkending:
Closing as issue will be fixed by this issue https://github.com/choonkending/react-webpack-node/issues/107

Issue #57, opened 2015-08-20
User state after login doesn't contain req.user data


joshhornby:
Since the new React Router PR (https://github.com/choonkending/react-webpack-node/pull/52) I am doing something like the following once a user has entered the correct details https://github.com/rackt/react-router/blob/master/examples/auth-flow/app.js#L80 

Once they get redirected the user store doesn't contain the server user state which can be seen here https://github.com/choonkending/react-webpack-node/blob/master/server/config/routes.js#L62. The user state only contains what has been set at https://github.com/choonkending/react-webpack-node/blob/master/app/stores/UserStore.js#L67

Once I refresh the page the state contains the entire user object, this could be an issue with react router and the `transitionTo` but I am not sure.

choonkending:
@joshhornby Hey I had a look into this. You can easily pass in the req.user object to be bootstrapped by alt.

Below is a horrid example:
```javascript
// Naming convention leaves a lot to be desired
var user = req.user ? { object: req.user, authenticated: true, isWaiting: false } : { object: null, authenticated: false, isWaiting: false };
```

Whereas your Dashboard.jsx could be something like:

```javascript
import React from 'react';
import UserStore from 'stores/UserStore';

export default class Dashboard extends React.Component {
  constructor(props) {
    super(props);
    this.state = UserStore.getState();
  }

  componentDidMount() {
    UserStore.listen(this._onChange);
  }

  componentWillUnmount() {
    UserStore.unlisten(this._onChange);
  }

  _onChange = () => {
    this.setState({
      user: UserStore.getState().user
    });
  }
  render() {
    console.log(this.state.user.get('object'));
    return (
      <div>Welcome to the Dashboard
        <div>{this.state.user.get('object').get('email')}</div>
      </div>

    );
  }
}

```

We'll be using AltContainer soon hopefully so none of this listen to stores for every high level component lol

joshhornby:
@choonkending This still produces the same issue. After the intial login the user state only contains what has been set in the login action. Doing the example above doesn't work as `undefined` is returned when I do `props.user.get('object')` (I am already using AltContainer :+1: )

I will see if I can reproduce this bug this weekend and get an example repo online.

choonkending:
@joshhornby Can you also have a look at https://github.com/choonkending/react-webpack-node/pull/61? This is how I got AltContainer working.

joshhornby:
Sure. I'll base my example repo off this.

choonkending:
[...]

Issue #56, opened 2015-08-19
Layout switching


praveen-ti:
How to create a new layout for admin here. What all are the changes needed in routes and webpack.config files.

joshhornby:
@praveen-ti I have created a PR (#62) with outlines what you will need to do. The PR also has some new code with isn't relevant to what you need, so to outline:

1) Create a new class which will house the admin layout (https://github.com/choonkending/react-webpack-node/pull/62/files#diff-b9823c8ef410f4b456968a4046e79272R1)
2) In the `rotues.jsx` file you will want to wrap everything in a `<Route>` and then create a new block (https://github.com/choonkending/react-webpack-node/pull/62/files#diff-70012bb93dfce2be702a0251971c1db5R20)

Now in this case when the `login` route is hit, the layout will be taken from the `UnauthorisedApp` component and not the `App` component. 

Any questions let me know :+1:

praveen-ti:
@joshhornby Thank you for you reply. I have fixed the issue. Created new layout for admin and switched layouts based on routing.
const routes = (
  <Route>
    <Route name="appadmin" path="/admin" handler={Admin} >
        <Route name="admin" handler={Adminlogin} />
        <Route name="/admin/login" handler={Adminlogin} />
        <Route name="/admin/logout" handler={Adminlogout} />
        <Route name="/admin/languages" handler={Languages} />
        <Route name="/admin/pages" handler={Pages} />
        <Route name="/admin/dashboard" handler={Dashboard} />
        <DefaultRoute handler={Adminlogin} />
        <NotFoundRoute name="404page" handler={NotFound} />
    </Route>
    <Route name="app" path="/" handler={App} >
      <Route name="login" handler={Login} />
      <Route name="logout" handler={Logout} />
      <Route name="index" handler={Index} />
      <Route name="signup" handler={Signup} />
      <Route name="signup/:hash" handler={Signup} />
      <DefaultRoute handler={Index} />
      <NotFoundRoute name="404" handler={NotFound} />
    </Route>
  </Route>
);

praveen-ti:
Hi,
Thank you for your reply.  I have fixed the issue.




Regards,Praveen

choonkending:
@joshhornby Too quick!

PR #62, opened 2015-08-22
Separate layouts & bug showing req.user object

joshhornby:
# Related Issues

#56 #57 #58 

This PR shows two things

1) The ability to split the login route into a different 'layout' - this will allow theming etc.
2) An example showing the issues stated in issue #57 - Once you login you will be auto redirected to the dashboard (https://github.com/choonkending/react-webpack-node/compare/master...joshhornby:feature/layout?expand=1#diff-129f1ad504e80d839c390c823d8ad3c2R70) You will notice you should see the users email in the nav bar (https://github.com/choonkending/react-webpack-node/compare/master...joshhornby:feature/layout?expand=1#diff-129f1ad504e80d839c390c823d8ad3c2R70) - this has been added to the `user` object in the server routes file (https://github.com/choonkending/react-webpack-node/compare/master...joshhornby:feature/layout?expand=1#diff-99de27cb2ded592bf6ae68b3d132eaceR62) But the email is only shown AFTER a refresh of the page - this is because after the redirect the user state only contains the following 

`Object {authenticated: true, isWaiting: false}`

Which is set here https://github.com/choonkending/react-webpack-node/compare/master...joshhornby:feature/layout?expand=1#diff-129f1ad504e80d839c390c823d8ad3c2R67

# Todos

- [ ] I don't love the way the router is passed into the store - it just doesn't feel correct
- [ ] Make the redirect dynamic as seen in this following example - https://github.com/rackt/react-router/blob/master/examples/auth-flow/app.js#L79

@choonkending Looking forward to hearing your feedback, especially regarding point 2. This may also be related to issue #58 as that should handle the redirect for us come to think of it, but as outlined on there it doesn't always seem to be fired.

### DIFF
--- a/app/actions/UserActions.js
+++ b/app/actions/UserActions.js
@@ -11,19 +11,22 @@ class UserActions {
 
   manuallogin(data) {
     this.dispatch();
-    UserWebAPIUtils.manuallogin(data)
+
+    let { email, password, router } = data;
+
+    UserWebAPIUtils.manuallogin({email: email, password: password})
       .then((response, textStatus) => {
         if (textStatus === 'success') {
           // Dispatch another event for successful login
-          this.actions.loginsuccess(data.email);
+          this.actions.loginsuccess(router);
         }
       }, () => {
         // Dispatch another event for a bad login
       });
   }
 
-  loginsuccess(email) {
-    this.dispatch(email);
+  loginsuccess(router) {
+    this.dispatch(router);
   }
 
   // Leaving this here for future use
--- a/app/components/Login.jsx
+++ b/app/components/Login.jsx
@@ -36,7 +36,8 @@ export default class Login extends React.Component {
     const password = React.findDOMNode(this.refs.password).value;
     UserActions.manuallogin({
       email: email,
-      password: password
+      password: password,
+      router: this.context.router
     });
   }
 
@@ -74,3 +75,4 @@ export default class Login extends React.Component {
 }
 
 Login.propTypes = { user: React.PropTypes.instanceOf(Immutable.Map) };
+Login.contextTypes = { router: React.PropTypes.object };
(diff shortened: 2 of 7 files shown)


choonkending:
@joshhornby Awesome work! I like the way you separated `UnauthorisedApp`, seems really reasonable.

For point 2, I think I know what might be causing it, but I gotta pull this and test it and get back to you 

 > I don't love the way the router is passed into the store - it just doesn't feel correct

Agreed. Feels wrong.

> Make the redirect dynamic as seen in this following example

Yep! This is a good one

choonkending:
@joshhornby I have not fixed issue two, but I did find a better way to not use router in the store.

```javascript
// UserActions.js
  manuallogin(data, cb) {
    this.dispatch();

    let { email, password, router } = data;

    UserWebAPIUtils.manuallogin({email: email, password: password})
      .then((response, textStatus) => {
        if (textStatus === 'success') {
          // Dispatch another event for successful login
          this.actions.loginsuccess(email);
          // Execute callback function after done
          cb(true);
        }
      }, () => {
        // Dispatch another event for a bad login
        cb(false);
      });
  }

```

```javascript
// Login.jsx
  _onLoginSubmit = () => {
    const email = React.findDOMNode(this.refs.email).value;
    const password = React.findDOMNode(this.refs.password).value;
    UserActions.manuallogin({
      email: email,
      password: password
    }, (loggedIn) => {
      if (loggedIn) {
        this.context.router.replaceWith('/dashboard');
      }
    });
  }

```

```javascript
// UserStore
  handleLoginSuccess(email) {
    this.user = this.user.merge({ isWaiting: false, authenticated: true, email: email});
    this.emitChange();
  }
```

joshhornby:
@choonkending Agreed this is certainly a nicer solution! I will update the PR

joshhornby:
Agreed. Exciting seeing Redux on here.

PR #67, opened 2015-09-02
Fixed broken webpack config link

mgmcdermott:
Noticed this trying to click from the npm site. Link to webpack config file is broken.


choonkending:
@mgmcdermott Good catch! Thanks a lot!

PR #83, opened 2015-10-19
Fix link to Alt Container

dbburgess:
Very minor issue, but the link was broken, so I found a relevant link
and updated it.


choonkending:
Thanks @dbburgess! Good pickup!

Issue #82, opened 2015-10-15
Update to React 0.14!


choonkending:

bishopZ:
I made a boilerplate based from yours and did upgrade it to React 0.14. There were only a few changes which were covered pretty well in the upgrade guide. If my repo helps as a reference for this or anything else: https://github.com/bishopZ/kabuki-motion

choonkending:
@bishopZ Thanks for that. I've made the changes in a https://github.com/choonkending/react-webpack-node/pull/84, just have not sorted out the extra warnings in console with AltContainer yet :D

bishopZ:
Yeah, I ended up removing AltContainer in the short term. I was able to get
everything working by just importing the Store where necessary, fwiw.

On Wed, Oct 28, 2015 at 10:44 AM, Ken Ding <notifications@github.com> wrote:

> @bishopZ <https://github.com/bishopZ> Thanks for that. I've made the
> changes in a #84
> <https://github.com/choonkending/react-webpack-node/pull/84>, just have
> not sorted out the extra warnings in console with AltContainer yet :D
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/choonkending/react-webpack-node/issues/82#issuecomment-151887751>
> .
>



-- 
________________________________________________________________
================================================================
=--------------------------------------------------------=-===-=-=-====--
+_+~_=_~--+__+=-^=-+_+_=^-+__+-=+_+~__=__~-_-____-=++=_--^-===-=-==-=-=--
=--------------------------------------------------------=-===-=-=-====--

http://bishopZ.com
_______________________________________________________________________

choonkending:
Ready to merge

PR #84, opened 2015-10-20
react 0.14

choonkending:
Still need to remove warnings in console


choonkending:
https://github.com/choonkending/react-webpack-node/issues/82

PR #89, opened 2015-11-05
Migrating to Redux

hsin421:
In this PR:
1. Migrate from Alt to Redux with Redux, React-Redux, Redux-Router (Redux + React-Router), Redux-thunk and Redux-devtools. 

2. Remove Immutable (need help set up Redux + Immutable). 

3. It still needs minor fixes, mainly export syntax errors in redux/actions and some Eslint breakages.

4. Protected route (that needs authentication) needs works, as Redux-router doesn't play well with onEnter hook in React-router.


choonkending:
@hsin421 Awesome stuff man!

hsin421:
Can you take a look at the export structure in redux/actions/user.js or topic.js. It was giving me errors the way I did it. 

It works on my previously cloned copy, but not at this current branch. Did you add new Eslint rules or new Webpack configs? Or it may be 0.14, as I was still on 0.13.

choonkending:
@hsin421 Sure I'll take a look at it. Thanks for your hardwork.

I'm not sure which version of the cloned copy you were using. We've been adding some eslint rules + webpack config. 0.14 is in a separate branch not merged yet :)

choonkending:
@hsin421 Do you mind merging the latest into this PR? (I've sneakily added react14 - it's been a long time waiting).

choonkending:
@hsin421 The code looks good. I'm happy to help you debug this once you push an updated PR up. Anxious to get redux out! :D

hsin421:
@choonkending roger that!

joshhornby:
<3

bellatrix1001:
@hsin421 Just curious.. why do you advocate migrating to Redux? I am pretty comfortable with alt (thanks to @choonkending) and love using their container approach, but it does look like development has slowed down a bit...

kkotwal94:
Gotta ride the hype train bro!

hsin421:
@bellatrix1001 When your app is simple it really makes no difference with Alt or Redux. But when your app grows beyond 3 stores than it is really cumbersome to keep track of them. That's where the single Store philosophy of Redux comes in handy. Plus Redux devtools are great and help speed up development. No more console logging all over to see if a action is firing.

choonkending:
@bellatrix1001 Thanks for raising that as a concern. I think alt is definitely a good starting point and it can scale. The reason for me to consider redux is because of what @hsin421 mentioned, it is easier to reason about. Also I believe redux has a good ecosystem and tremendous support as well.

As for the container approach from alt, I believe there is an equivalent to it http://rackt.org/redux/docs/basics/UsageWithReact.html (typing on the train so forgive the bad formatting)

I'd be happy create a separate repository for redux (I already have one empty repo for this) and continue maintaining this repo if many are happy with the alt implementation and do not want to switch.

SOSANA:
@choonkending 1+ 
I agree with you, create a new branch with redux, and another new branch in future when the next hottest flux implementation comes alone (cough cough relay).

Heck I think we should have a branch for reflux as well ;)

kkotwal94:
I think @SOSANA 's idea is pretty good

hsin421:
@SOSANA Redux is not strictly Flux, though.
For me at least I feel the community is converging toward Redux. Here're some of the reasons why from the Redux author Dan himself:
"Redux is not that different from Flux. Overall it's the same architecture, but Redux is able to cut some complexity corners by using functional composition where Flux uses callback registration.

It's not fundamentally different, but I find it that Redux makes certain abstractions easier, or at least possible to implement, that would be hard or impossible to implement in Flux."

Read more here:
http://stackoverflow.com/questions/32461229/why-use-redux-over-facebook-flux

choonkending:
@SOSANA I think having separate branches for flux implementations (loosely including redux here, as @hsin421  pointed out it s not strictly Flux) is a good idea, however, it might turn into a maintenance if we do not reign in the scope here. 

## Approach 1 
- alt in master branch
- Have redux as a separate branch that is actively maintained (it could diverge very easily from the master branch pretty quickly)

## Approach 2 
- redux as master branch
- alt in separate branch. Same problem as Approach 1

## Approach 3 
- Different repository for redux

## Approach 4 
- Nuke alt with breaking changes into redux
- Whatever future implementations (relay) will follow suit
- I'm interested in relay but haven't looked much into it

Hahaha. This is gonna be interesting. I'm happy to go with any approach, but 3 and 4 are what appeal to me.

kkotwal94:
I say approach 4 for the win

app/redux/actions/topics.js
```diff
@@ -0,0 +1,74 @@
...
+		dispatch(create(data));
+		return TopicWebAPIUtils.addTopic(data)
+		.done((response) => {
+			// Dispatch another event for successful login
+		}) 
+		.fail(err => {
+        // dispatch an event if fails to notify user that it has failed
+    });
+	};
+}
```
choonkending: @hsin421 I think there s a missing `}` here :D

app/redux/actions/topics.js
```diff
@@ -0,0 +1,74 @@
+import * as types from '../constants/actionTypes';
+import TopicWebAPIUtils from 'utils/RestaurantWebAPIUtils';
```
choonkending: :) Think we need don't have a RestaurantWebAPIUtils

app/redux/store/configureStore.js
```diff
@@ -0,0 +1,39 @@
+import { createStore, compose, applyMiddleware } from 'redux';
+import rootReducer from '../reducers';
+import { reduxReactRouter } from 'redux-router';
+import createHistory from 'history/lib/createBrowserHistory';
+import routes from 'routes.jsx';
+import thunk from 'redux-thunk';
+import createLogger from 'redux-logger';
+import { devTools } from 'redux-devtools';
+import { canUseDOM } from '../../../node_modules/react/lib/ExecutionEnvironment';
```
choonkending: This one was error-ing out for me.

choonkending:
@hsin421 This is some really good work. A couple of minor bugs remaining to be fixed (will comment) later, but this PR is looking good to be merged soon! Thanks a lot!!

choonkending:
I'm thinking of going with approach 4, but perhaps making the alt a separate branch so people still can refer to it. We will need to write a good migration guide from alt to redux as well.

app/redux/actions/topics.js
hsin421: Darn it, I thought I was fast enough that you didn't see this! :P

hsin421:
By the way, I really lament getting rid of Immutable in this PR. 

It would really help if someone fluent in Immutable can add it back in in a future PR.

choonkending:
@hsin421 Agreed. 

I reckon we adopt an approach like react-router. 

1. I'll make the current master branch `feature/alt` and make it the default branch
2. I'll merge this PR into master once we've got an MVP out.
3. We work on master for redux.

package.json
```diff
@@ -62,8 +66,12 @@
     "passport-local": "^1.0.0",
     "react": "^0.14.2",
     "react-dom": "^0.14.2",
-    "react-helmet": "^1.0.1",
+    "react-helmet": "2.1.1",
```
choonkending: This is a good change. But some of the changes they've bumped up require us to have a little change in `base.html`

```javascript
<!doctype html>
<html lang="">

<head>
    TITLE

    META

    LINK

</head>

```

choonkending:
@hsin421 Good work!

Final comments:

There was an issue with updating the counts on the server, because the index sent to the server was not the index created by the date stamp (it was the index in the Array). I think you've already partially implemented a fix for this, which was to add an additional `index` argument into the `incrementCount` function, however you will need to add that as a prop to `MainSection` component.

```javascript
 export default class MainSection extends React.Component {
   render() {
     const topics = this.props.topics ? this.props.topics.map((topic, key) => {
      return (<TopicItem index={key} id={topic.id} key={key} text={topic.text} />);
     })
     : null;
     return (

```

also,

```javascript
 class TopicItem extends React.Component {
   _onIncrement = () => {
    this.props.dispatch(incrementCount(this.props.id, this.props.index));
   }
 
   _onDecrement = () => {
    this.props.dispatch(decrementCount(this.props.id, this.props.index));
   }
 
   _onDestroyClick = () => {
    this.props.dispatch(destroyTopic(this.props.id, this.props.index));
   }

```

With that I'm happy to merge. Lots of docs to update. Haha!

The solution is definitely a good initial iteration towards moving towards redux (we probably can think of a more elegant approach once we get this out). Thanks so much for the hard work!

choonkending:
For reference:
1. Make redux folder the main folder (we don't need to have duplicate stores)
2. Perhaps add immutable
3. Clean up unneeded code

choonkending:
Created a new branch for the alt implementation. And possibly change ALL the readmes hahaha. 

https://github.com/choonkending/react-webpack-node/tree/flux/alt

choonkending:
@hsin421 Any chance you get the changes out soon (I'm preparing to give a talk about this at a meetup) :P

choonkending:
@hsin421 and all, thanks so much for your hard work!

hsin421:
Oops and thanks!! Glad to see it merged!

hsin421:
By the way, there's a formatting issue on the doc regarding Alt/Redux:

"On master branch:
Redux on Alt/Flux branch:
Alt...."

kaizvn:
massive PR ever lol.

PR #91, opened 2015-11-13
res.send() depricated and update css-loader

thientran1707:
- res.send(status, body) is deprecated. Change to res.status().send() instead.
- update css-loader to avoid deprecated options.
- react-router 1.0.0 is released

### DIFF
--- a/app/server.jsx
+++ b/app/server.jsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import ReactDOMServer from 'react-dom/server';
-import { RoutingContext, match } from 'react-router'
+import { RoutingContext, match } from 'react-router';
 import createLocation from 'history/lib/createLocation';
 import Iso from 'iso';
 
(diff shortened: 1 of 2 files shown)


choonkending:
@thientran1707 This is awesome mate. I'll test and merge this later if it passes. Thanks a lot!

thientran1707:
do you think it's a good idea to create node es6?

thientran1707:
in server.jsx, const renderToMarkup = (alt, state, req, res) => {}
i think we can remove alt as attribute, because you import alt already. Please correct if i am wrong :)

choonkending:
@thientran1707 Ah good question. It's because we need to pass the context of that particular alt instance  *after* it's been webpack-ed. The function that gets called still remembers the alt state and manages to bootstrap the stores. :)

(Alt will be changing soon, we will be moving to Redux as a main branch).

thientran1707:
oh ok, I got it. Btw, are you going to change the Node server to es6?

choonkending:
@thientran1707 Yes. Definitely. I actually run it locally with node4 (haven't upgraded to nodev5 yet).

We can change the package.json's `engine` to 4.2.1 and give it a go.

Also, we need to change package.json to have history@1.12.0

Got the following error message when compiling.
 react-router@1.0.0 wants history@^1.12.0

choonkending:
@thientran1707 Hey we'll be using redux for our master branch from now onwards, do you mind submitting to PR to flux/alt branch? :)

thientran1707:
it's ok :) actually i also would like to submit it to redux.haha

choonkending:
@thientran1707 I will close this PR in favor of https://github.com/choonkending/react-webpack-node/pull/97.

PR #105, opened 2015-12-27
upgrade to babel 6

kaizvn:
- Config to make project runnable on babel6.


kaizvn:
added staging-0 back as original `.babelrc`

PR #106, opened 2015-12-27
Upgrade to babel 6

kaizvn:
- Config to make project runnable on babel6.
- Added staging-0 back as original `.babelrc`


choonkending:
@kaizvn Thanks a lot! I'll have a look!

choonkending:
@kaizvn Haven't begun looking yet actually. Want to focus on getting my PR for unit testing https://github.com/choonkending/react-webpack-node/pull/102 merged so stay tuned! Haha!

kaizvn:
@choonkending I know it's OK, just leave it here till the  babel 6 support `decorator`, then we'll take care this one :joy:

choonkending:
@kaizvn Don't worry, it was a very good attempt! I'll have a look when I have the chance and we can see how whether we start a new PR or use this one.

.babelrc
```diff
@@ -1,20 +1,25 @@
 {
-  "stage": 0,
+  "presets": [
```
choonkending: I reckon we can follow the presets specified [here](https://github.com/gaearon/react-transform-boilerplate/blob/master/.babelrc). No issue with having stage-0. But we could look into using `react-hmre`.

server/config/routes.js
```diff
@@ -54,7 +54,7 @@ module.exports = function(app, passport) {
   // fetched and seed our stores with data.
   // App is a function that requires store data and url to initialize and return the React-rendered html string
   app.get('*', function (req, res, next) {
-    App(req, res);
+    App.default(req, res);
```
choonkending: Interesting. http://stackoverflow.com/questions/33505992/babel-6-changes-how-it-exports-default. Babel 6 changes the way we export functions. Will look into this and let you know how I think we can fix this. Not a fan of `App.default`.

kaizvn: @choonkending  agree. I dislike it too, I intend to put it on `var App = require('../../public/assets/app.server').default;` but it's unsafe. Still waiting for better solution.

choonkending: @kaizvn Good call. 

Perhaps we could change to using 
```
module.exports = function render() {}
```
in server.jsx

kaizvn: @choonkending let wait till babel6 release the new version that we could do merge, then we'll discuss about this :smile:

choonkending: @kaizvn Are they likely to be changing it in babel 6? Perhaps the previous behaviour we got used to in ES5 was wrong. Haha.

choonkending: @kaizvn We could either roll with the existing solution you have, or try [this](http://stackoverflow.com/a/33683495/1710968). I'm happy with either, as long as we document it saying this is `expected behaviour is ES6` with a link

choonkending:
@kaizvn I am happy to merge but waiting on the `.babelrc` file change, changing of presets :)

choonkending:
@kaizvn I reckon we should proceed with this. :)

kaizvn:
@choonkending so do we need to make any change with `.babelrc`?

choonkending:
@kaizvn I reckon we should :D :D

choonkending:
Closing, we have https://github.com/choonkending/react-webpack-node/pull/153/files . Thanks for that, I used this as a reference.

Issue #114, opened 2016-01-01
react-router 2.0


choonkending:
https://github.com/rackt/react-router/blob/master/upgrade-guides/v2.0.0.md

kaizvn:
@choonkending  just note about `history`, before it was included in NodeJS platform but now it separated already. So remember `npm install --save history` to avoid `undefined error` ( I've experienced with this :joy: ).

choonkending:
@kaizvn Thanks a lot! Will do :)

choonkending:
https://github.com/rackt/react-router/releases still using as release candidates. Will upgrade once it's out of rc (which I expect will be quite soon) :)

Issue #107, opened 2015-12-29
Better auth flow for react-router


choonkending:
# EDITED

## Goal of this issue fix
We want a good example of an authentication flow for universal redux and react-router. This will include:
- [x] Redirecting users to the Login page when they access a route/page which requires authentication.
    - [x] Server-side authentication redirects when the user accesses a route directly, e.g. `/dashboard` 
    - [x] Client-side redirects

# Scenarios

## Client side redirect
* Given I am not logged in
* And I am in the `/` route
* When I navigate to `/dashboard` which requires authentication
* Then I will be redirected to the `/login` path

## Server side redirect
* Given I am not logged in
* When I navigate to `/dashboard` which requires authentication
* Then I will be served the `/login` page

Possible example implementations:
https://gist.github.com/iNikNik/1eabfc6bcc132384368c

choonkending:
https://github.com/rackt/react-router/issues/2387

choonkending:
http://stackoverflow.com/questions/33358249/server-side-auth-flow-with-react-router?rq=1

hoodsy:
@choonkending have you done any work on this? I'm looking to implement this in a personal project, and would love to help with / take a look at your implementation.

choonkending:
@hoodsy I've been looking at different solutions. 

The main problem is dealing with server-side redirects (when a user is not authenticated) in a nice and understandable way. 

Really happy to get help/suggestions with tightening up the auth flow! Thanks a lot! So far https://gist.github.com/iNikNik/1eabfc6bcc132384368c looks to be a good one. 

I feel like I'm missing something simple/there could be a simple solution to this.

hoodsy:
@choonkending I've been doing a bit of research as well, and liked the ```onEnter``` approach shown in that gist best. 

What do you think about only redirecting with redux-simple-router, for consistency sake? Using  ```pushPath``` all app routing could happen from the front end. 

I was also thinking that  ```onEnter``` would fire off an action to test  ```user.token```. The token, received from the server, could be persisted via localStorage (or redis) to maintain sessions. 

So the flow would be something like this:
Login -> auth action -> server responds with token -> token stored in user -> onEnter action called for route change -> if user.token, gain access -> request to server for route data (carrying token) -> server response, if token is valid, receive route data 

I hope this is clear, and it would be great to hear your thoughts!

choonkending:
@hoodsy Sorry for the late reply.

[...]

PR #125, opened 2016-01-20
Feature/auth flow

choonkending:
- Upgrade to react-router 2.0, redux-simple-router 2.0
https://github.com/choonkending/react-webpack-node/issues/114
- Auth flow
https://github.com/choonkending/react-webpack-node/issues/107


hoodsy:
@choonkending the only reason I could see for the ```req.isAuthenticated()``` always returning false is the session cookie not correctly being stored.. but I'm not sure why this would be the case.

choonkending:
@hoodsy I know! I initially thought I was not sending cookies because of isomorphic-fetch, but I have all the proper arguments specified in the fetch request. I'll have a look again later with a clearer mind.

choonkending:
I can verify that cookies are not sent from server fetch. Possibly because you shouldn't need to be checking whether you are authenticated via API if you're on the same-origin. 

Anyway, I think I have a solution from this brilliant repo. https://github.com/erikras/react-redux-universal-hot-example/blob/c22755e52565f0bfc829d88d31cb92b9749b8145/src/routes.js#L16

Testing it now before pushing it for you to have a go at it.

choonkending:
@hoodsy Have a go at it. I'll be merging it in soon-ish :)

Still need to solve the issue of redirecting *after* logging in, but I think that can be another branch.

choonkending:
@psimyn The long awaited fix.

hoodsy:
@choonkending looks great man! excited to get this into my project as well :+1:

PR #138, opened 2016-02-12
refactoring serverside async logic

charlieclark:
I've also swapped out `isomorphic fetch` +` es6 promise polyfill` for `axios`, which is a better (simpler) lib.
Take a look and let me know what you think!


choonkending:
@charlieclark Thanks very much for this! Will go through the code tomorrow! :+1: :smile:

charlieclark:
Oh yeah one weird thing I encountered, `import * as types from 'constants` was acting weird and importing a bunch of constants from a node_module I think. 

I didn't spend much time on it and just renamed the import to `contstants/index`. Worth looking into though. It does bring up the debate of whether absolute imports for user code is a good idea.

choonkending:
@charlieclark Sorry about that, I just merged a PR for linting (as it was just linting errors). I wanted to go through this PR more thoroughly, so could you pull from master again?

> import * as types from 'constants

I thought that was weird when I saw that you changed it to constants/index. Will have a look now.

app/actions/topics.js
```diff
@@ -1,10 +1,11 @@
 // Including es6-promise so isomorphic fetch will work
-import { polyfill } from 'es6-promise';
-import fetch from 'isomorphic-fetch';
+// import { polyfill } from 'es6-promise';
+// import fetch from 'isomorphic-fetch';
+import request from 'axios';
```
choonkending: No gripes with using `axios` as our library, as it looks to have better browser support and is well maintained. However, we might still have to use a promise polyfill for browsers an environment that does not support es6-promises. Found that [here](https://github.com/mzabriskie/axios#promises). 

Apart from that I'm happy for you to remove the commented lines.

app/api/fetchComponentDataBeforeRender.js
```diff
@@ -0,0 +1,14 @@
...
+
+export function fetchComponentDataBeforeRender(dispatch, components, params) {
+  const needs = components.reduce( (prev, current) => {
+    return (current.need || [])
+      .concat((current.WrappedComponent ? current.WrappedComponent.need : []) || [])
+      .concat(prev);
+    }, []);
+    const promises = needs.map(need => dispatch(need()));
+    return Promise.all(promises);
+}
```
choonkending: @charlieclark The code makes me so so so happy.

app/api/promiseMiddleware.js
```diff
@@ -0,0 +1,22 @@
...
+        next({ ...rest, req, type: SUCCESS });
+        return true;
+      })
+      .catch(error => {
+        next({ ...rest, error, type: FAILURE });
+        console.log(error);
+        return false;
+      });
+   };
+}
```
choonkending: :+1:

choonkending:
Excellent work!!

I really like the promise middleware approach + your fetchData from component approach :+1: for your reduce function :)

Couple of comments:
1. Using axios requires us to pass data as config I believe. It works for increment, decrement, but for destroy topics, we need to convert `{id:id}` into `{data:{id:id}}`
```javascript
{data: { [Object] }}

```
2. We will need to update the tests [ can be fixed after the PR is merged]
3. Unify all requests to use the promise middleware to avoid confusion [can be fixed after PR is merged]

I'm really happy with how this is turning out. I'm happy to merge once we fix 1. and pull master to resolve linting conflicts.

app/api/promiseMiddleware.js
choonkending: Is it worth moving this to a folder called `/middlewares` in case we wish to add more middlewares in the future?

app/actions/topics.js
charlieclark: good catch

charlieclark:
Ok I've made the updates.

I looked into why the destroy call wasn't working. One thing I noticed is that your implementation of the REST API is a bit atypical, and standard REST apis typically don't expect a payload with on remove/delete ( which is why axios doesn't allow data to be sent via the destroy/remove method).

I made a few updates, let me know what you think.

choonkending:
Very good!

> One thing I noticed is that your implementation of the REST API is a bit atypical, and standard REST apis typically don't expect a payload with on remove/delete ( which is why axios doesn't allow data to be sent via the destroy/remove method).

Good pick up.

Will be merging in a bit!

choonkending:
@charlieclark Thanks! :)

I will fix the tests (as I m planning to slowly increase test coverage) + give this a version bump! :D

Issue #150, opened 2016-02-27
postcss


choonkending:
(Edited 6/03/2016)

Here's a brain dump of why I think we should go with [PostCSS](https://github.com/postcss/postcss#postcss---) completely - and eliminating sass from our boilerplate. They are not directly comparable (1:1 match) I know, but since we already use autoprefixer as part of our webpack config, I reckon it's high time we consider removing sass to see if it works swimmingly well.

**tl;dr**

PostCSS is a tool for transforming styles with JS plugins. These plugins can lint your CSS, support variables and mixins, transpile future CSS syntax, inline images, and more.

**Pros**
- PostCSS is a **blank slate** - we have control over which features we want. 

Preprocessors such as sass ship a **monolithic** set of features (which are awesome), but just like how the JS ecosystem enables us to plug and play, I think using PostCSS encourages us to think "Do we need this? Is this really solving the problem?" or at the very least have a conversation around features, rather than bulldozing through css. We can even write plugins for postcss if we find the need to.
- **FAST**. Look at [this](https://ai.github.io/about-postcss/en/?full#42) slide. 

**Cons**
- **Deciding** + a mini learning curve

There will be some thinking involved when deciding *what* features we actually want when writing css.

- PostCSS is not a silver bullet. 

It does not replace everything Sass has, we might be able to achieve some feature parity with some plugins but it might not be what everyone likes.

**Considerations**
1. How this affects usage within the boilerplate?
There are a few features I think we cannot live without, that I plan to include right from the get-go:
- [postcss-simple-vars](https://github.com/postcss/postcss-simple-vars) for Sass-like variables. 
```css
$black: #000;
.someClass {
   color: $black;
}
```

Eventually move towards using [cssnext](https://github.com/MoOx/postcss-cssnext). Why not use features that will eventually be in our browsers?
- [postcss-nested](https://github.com/postcss/postcss-nested) for nested css rules
```css
.dream {
    &__in-a-dream {
          transform: rotate(180deg);
     }
}
```
- [PostCSS mixins](https://github.com/postcss/postcss-mixins) for mixins. Although with CSS modules, we could use `composes` if we needed to just reuse modular css without logic. 

The `@define-mixin` syntax differs a bit from Sass' `@mixin` keyword, but I'm sure we'll get over it soon enough.

```css
@define-mixin button $color {
    background: $color;
    padding: 12px 8px;
}

.button {
  @mixin button;
}
```
[...]

PR #154, opened 2016-03-03
Swap to using postcss instead of autoprefixer

dzotokan:
Addresses #150 

Not sure if it's everything you need to do to swap.


choonkending:
@dzotokan :+1:  Definitely first step to removing that annoying deprecated warning. 

Commented more on the issue.

Issue #157, opened 2016-03-05
Redux webpack console error in production


dzotokan:
Running the app in 'production' (`npm start`) throws the following error:

```bash
You are currently using minified code outside of NODE_ENV === 'production'. 
This means that you are running a slower development build of Redux. 
You can use loose-envify (https://github.com/zertosh/loose-envify) for browserify or DefinePlugin for webpack (http://stackoverflow.com/questions/30030031) to ensure you have 
the correct code for your production build.
```

Applying [this](http://stackoverflow.com/questions/30030031) fixes it. 

Any objections on puling it in?

https://github.com/dzotokan/react-webpack-node/commit/35b5fe6af7dce548861c7b64c397b2dd6a6e67ba

choonkending:
:+1

I agree we should get rid of that warning. But I found problems after deploying to Heroku - server side async calls require absolute urls, and we require a `process.env` hostname and port, which define plugin overrides when setting process.env.

Might be doing server side async calls wrong, I'll have a look to see how we can fix this !

choonkending:
Found a fix for this (from this [tweet](https://twitter.com/tikotzky/status/679796396149596160))

https://github.com/tikotzky/inline-environment-variables-webpack-plugin helps fix the problem. 

The problem here is that when using `DefinePlugin` for `process.env` to replace `NODE_ENV`, the `process.env.HOST_NAME` and `process.env.PORT` does not get set for Heroku :(

The reason why we need that is because of [this](https://github.com/choonkending/react-webpack-node/blob/master/app/server.jsx#L17) of code. For server-side async api calls (used for server-side rendering), we need to specify an absolute URL instead of using a relative `/api/topic` URL. Please correct if I'm wrong. I've inspected and found that using a relative path we get a `CONNECTION refused` error, accessing the wrong hostname and port.

After a bit of a google, the tweet saved me. So you won't see any errors in console for production any longer.

choonkending:
@dzotokan Let me know if that fixes it.

PR #160, opened 2016-03-05
Fix redux warnings by inlining process.env.NODE_ENV value

choonkending:
:boom:


choonkending:
Fixes https://github.com/choonkending/react-webpack-node/issues/157

PR #169, opened 2016-03-13
Added Windows support using cross-env library

Magneticmagnum:
I added the cross-env as a dependency and to the scripts.  cross-env provides a script which sets an environment unix style to work on Windows and linux operating systems.

I don't have a linux machine to test on.  Once the user gets MongoDB running on the default port, user can simply run 'npm install' then 'npm run start' and everything will work great with these changes on Windows 10 64-bit and Windows 7 64-bit on Node version 5.5.0 and NPM version 3.3.12.

These settings should not affect linux / mac users.  It would be great if someone can test on those machines.


choonkending:
@Magneticmagnum Thanks!! I'll get to testing on mac soon!

philohelp:
Yep I did that too : it's the kind of tricks a windows user must know. For a reason though, I had to specify the env manually before launching the app otherwise I got "environment=undefined". 

Just run the following command each time you want to change the environment : 

`set NODE_ENV=development` or `set NODE_ENV=production`

choonkending:
@Magneticmagnum Thanks for this. Merged!

Magneticmagnum:
No problem, great repo!

Issue #170, opened 2016-03-13
Including images in css


choonkending:
Raised by @dzotokan

PR #172, opened 2016-03-14
Feat/there can only be postcss

choonkending:
https://github.com/choonkending/react-webpack-node/issues/150

https://github.com/choonkending/react-webpack-node/issues/170


choonkending:
@psimyn Postcss + Css modules + values

choonkending:
Also logical commits so u could cherry pick to apply the webpack config

choonkending:
Will be merging by today

Issue #156, opened 2016-03-03
Slim & stateless version of the boilerplate


dzotokan:
Cloned the project and didn't need any of the auth/db/models things.

Any thoughts on stripping those out and maybe creating a separate branch?

i.e. Static react app with routes and an empty `need` maybe, with no authentication and DB. Would be more suitable in a microservices architecture land perhaps.

Happy to write up a PR removing those things (not for `master` obviously).

choonkending:
@dzotokan Definitely! 

This has been bothering me because it took me at least 30 minutes stripping it out that day. 

>  Static react app with routes and an empty need maybe, with no authentication and DB. Would be more suitable in a microservices architecture land perhaps.

Good thinking. Perhaps we only need to keep that branch up to date when there are webpack or config changes then.

choonkending:
I'm going to try to work on this guy. The refactoring here will mainly be in the `server/` folder.

The easiest approach I can think of is to:
- abstract connecting + authentication + model creation  into a different file - so people can easily comment that out if you wish to switch databases/not use a database.

I'll have a go to see how much work this will take + whether it's a good solution

choonkending:
Closing as issue is tacked in #372

Issue #94, opened 2015-11-19
When will you implement the Token Based Authentication?


sunkant:
This project is very good, but when will you implement the Token Based Authentication? : )

choonkending:
@Luke currently looking to tidying up and using Redux better and then
writing tests. When would you like this out?

If you have a solution for this already, happy to help out or check a PR :)

lukeqi:
Hi @choonkending Actually I am looking for a way to use Token based auth with Redux, but I encounter some problem and I am still working on it.

This project: https://github.com/nicolashery/example-isomorphic-one
Uses SSR, Token based Auth with Fluxible front framework, and I think this project shows a possible way to solve it.

lukeqi:
It load token from request cookie first every time when a page refresh request arrive on the server.

However, when I look at this project (which is very hot on github):
https://github.com/erikras/react-redux-universal-hot-example

It uses session based auth (as your project), and loads api from another server when doing SSR. This adds complexity when trying to do SSR if using token based auth.

lukeqi:
Like if we want to load api from another server (or different port in the same host, like localhost):

The "another" server need to obtain the Token from the cookie of the api Post,

So, when our SSR server issues the api Post, it must have something like a HTTP Interceptor, which adds the Token from the original request from the user.

Angular has such a thing like "HTTP Interceptor", but I do not find a similar one in Reactjs.

choonkending:
@lukeqi Sorry I took a while to get back on this. Yes, this looks interesting. I'll have to read up on token based authentication, but I think it's a good approach to the one we're currently having. :)

choonkending:
I think this has been sitting on the back burner for a while now. It should be the next thing we look at after abstracting DB. 

**Potential references**
1. [https://davidwalsh.name/react-authentication](https://davidwalsh.name/react-authentication)
2. [https://jwt.io/](https://jwt.io/)

mjewell:
Is anyone currently working on this? I'm happy to take a look at it

choonkending:
Go ahead! I don't think anyone is :)

Issue #26, opened 2015-06-22
ES6 Node


joshhornby:
Hey,

First of all love this application - it is a fantastic learning tool!

I have a quick question, is there a reason the node part of this app isn't wrote using ES6?

choonkending:
@joshhornby Thanks for your feedback!! :) 

I didn't rewrite the node part in ES6 initially because it was not part of the webpack build step and I thought Heroku didn't support the node 0.12 runtime. 

But I'm happy to have the node end have a ES6 refactor. If you're interested, would you like to submit a PR for this?

joshhornby:
I should have some time to ES6 the node side, although may need some help with the webpack build step.

choonkending:
@joshhornby Sure thing! Thanks!

SOSANA:
1+ be nice for people coming from ES5 to ES6 to see the difference. I will be hacking away with this. Be cool to have a branch for ES5 just for learning but understand the need to move forward with ES6.

choonkending:
@SOSANA Thanks!

Yea I agree. I considered having separate branches for ES5 and ES6 when I made the ES6 change, but went ahead with just one branch and thought having docs for it would be good enough. 

I have a small README [here](https://github.com/choonkending/react-webpack-node/blob/master/docs/ReactInES6.md) for moving to ES6 from ES5 with React. If you like you can create a separate ES5 branch, or update that doc :)

SOSANA:
@choonkending your so money, Just checked outed README. Good job! Hope to contribute when I can :)

choonkending:
I've been using io.js to run npm.

We could go with node v0.12 with the --harmony flag (i think)

or io.js (which supports ES6 by default)

 @joshhornby what do you think? I'm leaning towards io.js. 

Heroku also supports this: https://devcenter.heroku.com/articles/nodejs-support#specifying-a-node-js-version

joshhornby:
@choonkending I have no issues with switching to io.js.

Although whats the deal with io.js and node.js merging? I thought I heard discussions at one point?

choonkending:
I'm using node 5 currently. Just haven't made the switch for the server side code. Should really look into this.

caranicas:
the best example I have found using es6 on the server is the react starter kit from kriasoft. https://github.com/kriasoft/react-starter-kit

the way it uses webpack is and the build scripts is pretty different, so I am still trying to detangle it all, and figure out why the bable-loader isn't running on our server code.

[...]

PR #176, opened 2016-03-16
Abstract out database connection [ WIP ]

choonkending:
**WIP** :warning: 

https://github.com/choonkending/react-webpack-node/issues/156

Creating this PR to give visibility - it's not done yet


choonkending:
@dzotokan @mjewell 

This is the initial pass of how abstracted out database connections.

**Summary**
Create `appConfig` file which is a shared object
```javascript
// For MongoDB
DB_TYPE: process.env.DB_TYPE || 'MONGO'
```
There are a few files where configuration is dependent on DB, namely:
- `connect.js` that connects to the DB
- `express.js`  where the `session.store` object requires a `connect-mongo`
- `passport.js` where serialising and deserialising that accesses the DB
- `routes.js` where the routes that requires controllers that access the DB

So for `postgresql` or we could possiby change the `DB_TYPE`.

**Starting up a server without a DB connection**
Currently the quickest way would be to:
```javascript
// server/config/appConfig.js  
DB_TYPE: DB_TYPE.EMPTY
.
.
.
// server/index.js
/*
 * comment out the following lines
 */
require('./config/connect')();
require('./config/passport')(app);
require('./config/routes')(app);

```

The front end still looks and remains the same atm.

server/config/constants.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+  DB_TYPE: {
+    MONGO: 'MONGO',
+    POSTGRES: 'POSTGRES',
+    EMPTY: null
```
mjewell: I think this makes more sense as NONE: null

mjewell:
This sounds good to me for the most part. I think routes.js itself should be unaffected, but all the controllers will need to be changed based on the database.

Also, maybe we can use JSON Web Tokens so we can remove sessions entirely, which would mean express.js no longer depends on the database type.

server/config/constants.js
choonkending: Good one!

server/config/connect.js
```diff
@@ -0,0 +1,41 @@
...
+
+    // Register schema as mongoose model
+
+    var modelPath = path.join(__dirname, '..', 'models');
+    fs.readdirSync(modelPath).forEach(function(file) {
+      if(~file.indexOf('.js')) require(modelPath + '/' + file);
+    });
+  }
+};
+
```
mjewell: This file could get pretty big, maybe make connect/index.js which exposes all of these on an object, like so:

```
var DB_TYPE = require('./constants').DB_TYPE;

module.exports = {
  [DB_TYPE.MONGO]: require('./mongo'),
  ...,
  ...
};
```

then use it like this:

```
var appConfig = require('./appConfig');

require('./config/connect')[appConfig.DB_TYPE]();
```

server/config/express.js
```diff
@@ -61,24 +63,29 @@ module.exports = function (app, passport) {
...
       secure: false,
-    },
-    store: new MongoStore(
-      { 
+    }
+  };
+
+  if (appConfig.DB_TYPE === DB_TYPE.MONGO) {
+    var MongoStore =  require('connect-mongo')(session);
+    sess.store =  new MongoStore(
```
mjewell: And maybe similar for this:

```
store: require('./sessions')[appConfig.DB_TYPE]()
```

server/config/connect.js
mjewell: And actually, if you do this, it would make more sense to have the NONE type map to 'NONE' in the constants so you can check for undefined here and throw an error indicating that their env variable is not set to a valid value to avoid unexpected errors from typos, etc.

server/config/constants.js
mjewell: Also this should probably be DB_TYPES (plural), which also makes it less confusing when used with appConfig.DB_TYPE

choonkending:
@mjewell Thanks for the feedback. Will go through a second round and rebase hopefully tonight!

choonkending:
Change to be made:
`config/constants.js`:
- [ ] Rename `DB_TYPE` to `DB_TYPES` to avoid confusion
- [ ] Rename `EMPTY: null` to `NONE: 'NONE'` to check for undefined values 

`config/connect.js`
- [ ] Split into separate files for each DB and only expose the functions in `index.js`

`config/express.js`
- [ ] `store: require('./sessions')[appConfig.DB_TYPE]()`

**JWT**
Previously mentioned in https://github.com/choonkending/react-webpack-node/issues/94. Will give this a shot after this initial PR gets merged!

server/config/sessions/index.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  MONGO: require('./mongo'),
```
choonkending: @mjewell I couldn't `[DB_TYPES.MONGO] : require('./mongo')` as you suggested (was returning undefined).

Not sure if this is because we need to compile it with babel beforehand? [Example here](https://babeljs.io/repl/#?evaluate=true&presets=es2015&experimental=false&loose=false&spec=false&code=var%20DB_TYPES%20%3D%20%7B%0A%20%20%20%20MONGO%3A%20'MONGO'%2C%0A%20%20%20%20POSTGRES%3A%20'POSTGRES'%2C%0A%20%20%20%20NONE%3A%20'NONE'%0A%7D%3B%0A%0Avar%20a%20%3D%20%7B%0A%20%20%5BDB_TYPES.MONGO%5D%3A%20'b'%0A%7D%3B%0Aconsole.log(a)) @psimyn Thoughts?

I did the simple MONGO just to get it over the line. If it s a matter of ES6, it might be worth kicking off #26

mjewell: Ah, sorry, didn't realize that was es6 specific syntax. You can always still do:

```
sessions = {}
sessions[DB_TYPES.MONGO] = require('./mongo);
...
module.exports = sessions;
```

I took a brief look at getting babel working on the server yesterday and it seemed fine in development, but haven't looked at precompilation for production yet. Maybe I can get something basic working tonight

choonkending: My bad - I derp-ed big time. 

It doesn't require ES6 syntax! I didn't `require` the correct object. 😑 

Garh. I pushed the changes.

choonkending:
Will be merging tomorrow morning!

PR #184, opened 2016-03-23
Universal axios middleware for safer server-to-server requests

Epochi:
This lets you use all 'static need' methods in initial rendering with authentication attached and without exposing actions that pass user sensitive data like cookies. 

I added two extra apis, modified middleware and had to change action promise, but only to the actions that want to be called by static methods.

ApiServer and ApiClient are pretty much identical,they're just functions that create custom axios instance, but I had to separate them so I could inject axios interceptor for cookies only on the server to server api calls.

The modified flow goes like this:

ApiServer is created inside 'app/server.js', on render. req.headers.cookie is given to it
ApiClient is created inside 'app/client.js'
These custom instances are passed into customized promiseMiddleware(it also has thunk merged into it), that way when fetchDataBeforeRender calls fetchTopics it gives a function that is constructed inside promise middleware with either ApiServer  or ApiClient 
```
export function fetchTopics() {
  return {
    type: types.GET_TOPICS,
    promise: (client) => client.get('/topic')
  }};
```
_the seeding _
```
promiseMiddleware(client)
const actionPromise = promise(client);
```

As far as I can tell this didn't break anything

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,30 @@
+Note: This is just a template, so feel free to use/remove the unnecessary things
+
+### Description
+- Type: Bug | Enhancement | Question
+- Related Issue: `#abc`
+
+---------------------------------------------------------------
+## Bug
+**Expected Behavior**
+
+**Actual Behavior**
+
+**Steps to Reproduce**
+
+----------------------------------------------------------------
+## Enhancement
+
+**Reason to enhance/problem with existing solution**
+
+**Suggested enhancement**
+
+**Pros**
+
+**Cons**
+
+-----------------------------------------------------------------
+
+## Question
+
+**How to?**


choonkending:
@Epochi Apologies for the **really** late reply to this PR. I appreciate the work you've put into this PR!

I'll have to go through this a bit slower than usual, mainly because I do not have much knowledge on how this works 😢 

In the meantime, do you mind rebasing with the current master?

Epochi:
@choonkending  hey, no worries, i'll try to do that this week.
I tried to document it as much as possible, so it shouldn't look too tangled up

18601673727:
Looks pretty good but too more comments inside the code, you may want to comment your propose via github's review functionalities @Epochi

I hope this got merged soon = )

app/api/ApiClient.js
```diff
@@ -0,0 +1,17 @@
+//create axios instance for client side usage only
+import axios from 'axios';
+
+
+function api(){
+  const client = axios.create({
+    //Set baseURL to what the user will be accessing
+    baseURL: `https://react-webpack-node.c9users.io`
```
choonkending: Might need to change this to a different baseURL?

app/api/ApiServer.js
```diff
@@ -0,0 +1,27 @@
+//This creates new axios instance that is only used by router for server-side api calls.
+//If user is authenticated we create an interceptor for all requests made from server to server that were instigated by client
+//It's basically only used for initial server side rendering
+import axios from 'axios';
+
+let clientConfig = {
+  host: process.env.HOSTNAME || '0.0.0.0',
+  port: process.env.PORT || '8080'
+};
```
choonkending: I think we might need to abstract this to a `clientConfig` file given that we're using it everywhere, but that should be done in a separate PR.

app/api/promiseMiddleware.js
```diff
@@ -1,26 +1,45 @@
 /*
 * Redux middleware to handle promises
-* As seen in: https://github.com/caljrimmer/isomorphic-redux-app
+* As seen in: https://github.com/erikras/react-redux-universal-hot-example/blob/master/src/redux/middleware/clientMiddleware.js
+* 
+* This passes all actions with 'promise' to be used by client, which is custom axios instance
+* Functions called by makeTopicRequest and makeUserRequest are working as intended, because they just don't accept 'client' as argument
 */
+<<<<<<< cfc092ce0e278ad34ed88c597dc39be8a5436836
```
choonkending: Might have to remove this bit 😄

app/api/promiseMiddleware.js
```diff
@@ -1,26 +1,45 @@
...
   return next => action => {
     const { promise, type, ...rest } = action;
+=======
+export default function promiseMiddleware(client) {
+  return ({dispatch, getState}) => {
+    return next => action => {
+      if (typeof action === 'function') {
+        return action(dispatch, getState);
+      }
+>>>>>>> universal components
```
choonkending: Also this guy

app/containers/Vote.jsx
```diff
@@ -88,4 +82,13 @@ function mapStateToProps(state) {
...
 // Read more about where to place `connect` here:
 // https://github.com/rackt/react-redux/issues/75#issuecomment-135436563
-export default connect(mapStateToProps)(Vote);
+
+const Vote = connect(mapStateToProps)(VoteContainer);
+
+//Data that needs to be called before rendering the component
+//This is used for server side rending via the fetchComponentDataBeforeRender() method
+//Moved the static 'need' to connect wrapper, so initial render doesn't duplicate calls 
+Vote.need =[
```
choonkending: We can safely move this guy back into the Component, given that https://github.com/choonkending/react-webpack-node/pull/201 has been merged 👍

app/api/promiseMiddleware.js
```diff
@@ -1,26 +1,45 @@
...
 
 export default function promiseMiddleware() {
   return next => action => {
     const { promise, type, ...rest } = action;
+=======
+export default function promiseMiddleware(client) {
+  return ({dispatch, getState}) => {
+    return next => action => {
+      if (typeof action === 'function') {
+        return action(dispatch, getState);
```
choonkending: Lol i had no idea redux-thunk was such a [small lib](https://github.com/gaearon/redux-thunk/blob/master/src/index.js#L4). 

I wish we could still use thunk separately from this promise middleware so that we are clear with the differences of each middleware's role.

choonkending:
@Epochi Thanks for the effort you put into this PR. It's really addressing an issue that's been bothering a lot of our users. 

I've commented on some of the bits that might need a bit of changes. Let me know if you need help with merging the changes with the latest master - happy to help!

It might be worth just `git cherry-pick <commit>` onto a new branch because there has been merge conflicts already and might be hard to sort out.

choonkending:
@18601673727 Also hopefully you could have a look at this PR as you have context to what needs to be done as well! Thanks heaps guys!

app/api/promiseMiddleware.js
Epochi: Yeah, sorry about these, it was first time i used git rebase, not really sure what happened here :D

choonkending: Not a problem! Rebasing always gets me! (especially when you do an interactive rebase lol)

app/api/ApiServer.js
Epochi: The reason I had to separate these out is because of how axios interceptors work. 

When you add an interceptor on axios instance it will get used on *all* of it's requests.

So it means I need to have two axios instances:

ApiServer one that gets initialized before configureStore in render function inside [server.js](https://github.com/Epochi/react-webpack-node/blob/master/app/server.jsx), and I put the interceptor on this axios instance, but client cannot interact with it. 

ApiClient instance gets created before configureStore inside [client.js](https://github.com/Epochi/react-webpack-node/blob/master/app/client.jsx). and works as expected. 

ApiServer lives only in server and gets used only for server to server.
ApiClient lives only in client and gets used only for client to server


I separated these out essentially because I'm scared that someone could somehow modify axios interceptors if they were created client side. But I don't know if this is possible.

Epochi:
@choonkending hey, I red through my code again and then #221. I think my goal could be achieved in few lines of code inside his project. I will try to contribute to @18601673727 with the  *Interceptors, the foundation of OAuth support* point.

choonkending:
@Epochi You can float the idea over here in code snippets if you like. :)

PR #187, opened 2016-04-05
Swap out MongoDB/Mongoose for Postgres/Sequelize

mjewell:


choonkending:
@mjewell Thanks for contributing! I had a quick look at this PR, good work!

Do you mind if I keep this PR open a while longer?

I'm still thinking how I could abstract our db connections to be simpler. I was working on https://github.com/choonkending/react-webpack-node/pull/176 before my holiday, so I'd like to take a look at it again. Once I make it easier, it should be an easy plug and play, I could potentially have a branch for each db connection.

mjewell:
Sure, however you want to handle it is fine.

I think abstracting that would be great, but how are you planning to deal with the variance between ORM APIs? Or would those be the different branches?

choonkending:
Currently I am thinking of making those different branches, and maybe use the yeoman generator we have to allow users to choose between different ORM.

In any case, I think this is definitely a good step in the right direction.

iiegor:
@choonkending, let me know when this is done, I'll update the generator with support for this if you want ;)

choonkending:
@iiegor sure thing!

Bartekus:
Perhaps what is really needed instead of ORM is really GraphQL... your stack pretty much rocks and decoupling it from db in the first place would just be pure win.
Personally I'm looking at dropping mongoose/mongo and replacing it with GraphQL/rethinkDB... In anyway, have a look at rethinkDB, I highly recommend it :D

choonkending:
Refer to https://github.com/choonkending/react-webpack-node/pull/193

PR #189, opened 2016-04-12
use babel-node for dev, and babel for prod so we can use es6

mjewell:
#26


choonkending:
@mjewell You're on fire! Reviewing now

webpack/webpack.config.prod.js
```diff
@@ -25,7 +25,7 @@ var commonLoaders = [
       ]
     },
     include: path.join(__dirname, '..', 'app'),
-    exclude: path.join(__dirname, '/node_modules/')
+    exclude: path.join(__dirname, '..', 'node_modules')
```
choonkending: Good catch

choonkending:
@mjewell Solid work - and also bug fixes to the webpack config along the way! 👍 👍 

I like that we have `babel-node` for dev and `babel` for prod.

I like that you re-introduced the `babelrc` file back into the repo, it was previously removed because of this [issue documented here](https://github.com/choonkending/react-webpack-node/pull/189/files#diff-9968b87917a60f452c8ff987566a38e7L19). I believe [this line](https://github.com/choonkending/react-webpack-node/pull/189/files#diff-6e23492dba99a9412295736f9aef4254L17) on the webpack-dev-client is the culprit. Only concern is that it might be confusing for users who see a `query` in webpack and also presets in `.babelrc`, so we could either:
- Document/comment in `.babelrc` explaining where the presets will be used
- Use command-line arguments to replace `babelrc

Apart from that, we'll need to get users to `rm -rf node_modules/` before `npm install` again because you'd get a `You have mistakenly installed the `babel` package, which is a no-op` error.

I'd like to merge this once I get your feedback on my comments. It should probably bump to 1.6 from here. 

Given that, I'd like to try to get https://github.com/choonkending/react-webpack-node/pull/176 merged in tomorrow (because I derp-ed on the `index` files, check out the comment).

Thanks for contributing!!

mjewell:
Ah ok, I think it makes sense to just have it in cmd line args then. I tried that initially but I had issues with it not correctly ignoring the folders I specified so I just got lazy, but I changed that back now. Only annoying thing is you have to duplicate it since it is used for babel and babel-node. I put the commands next to each other in package.json so hopefully it's clear to anyone that changes it that both probably need updating. 

Also, as far as I know they can just

```
npm uninstall babel
npm install babel-cli
```

which will be faster

webpack/webpack.config.dev-server.js
```diff
@@ -9,14 +9,14 @@ var commonLoaders = [
...
-    loader: 'babel',
+    loader: 'babel-loader',
     // Reason why we put this here instead of babelrc
     // https://github.com/gaearon/react-transform-hmr/issues/5#issuecomment-142313637
     query: {
       "presets": ["es2015", "react", "stage-0"]
     },
     include: path.join(__dirname, '..', 'app'),
-    exclude: path.join(__dirname, '/node_modules/')
+    exclude: path.join(__dirname, '..', 'node_modules')
```
alexjoverm: I was also wondering if this was an issue or if I was missing something... I see I was right, nice catch :+1:

alexjoverm:
Guys, I think now it can be removed the `/server/**` entry from the `.eslintignore` file, then we can have es6 linting in there, right?

choonkending:
@mjewell Awesome change! Do you mind re-basing with the latest changes from master? I hope to get this merged tonight.

mjewell:
Already done

choonkending:
Too quick! Thanks so much! I'll test this locally now before merging. Might not make it until tonight (depending on how long this train ride is) 🚅 .

choonkending:
Working and tested locally and in heroku. 

I'll merge this and add docs tonight to remind users to

```javascript
npm uninstall babel
npm install babel-cli
```

as you suggested.

I've always wanted to switch to es6 for server side but never got around to it. 
Thank you so much for contributing @mjewell !

PR #190, opened 2016-04-15
Swap databases with a single config change

mjewell:


mjewell:
Hey, I was thinking more about the way databases are used and it still seemed kinda awkward to add support for new databases, as well as to switch between them. I took a shot at getting it to work where all you have to change is the ```DB_TYPE``` env variable and the server will figure out what stuff it should support based on that. The main point is that now there is a DB folder in the server which contains all the database/ORM specific code, including:

- models
- controllers
- deserializing users
- connecting to the database
- session stores
- passport logic

As a result, if you want to add a new database like I did/would do for the Postgres change, all the associated code goes to a single place with no changes required in the main server logic. Additionally, the server can handle incremental changes: if you just want your new database to support users but not topics, just define ```controllers/users.js```. Don't care about google-oauth? Don't bother implementing ```passport/google.js```

choonkending:
@mjewell This is a very good abstraction.

> The main point is that now there is a DB folder in the server which contains all the database/ORM specific code,

I believe this is the way forward. I will pull and test and hopefully merge this in today. Thanks very much for this!

choonkending:
👍 I really appreciate your help with this change! Merged!

PR #193, opened 2016-04-18
Add postgres support with new db config setup

mjewell:


mjewell:
It was easier to just create a new one with all the changes, so you can close #187 and use this instead. I added a separate sequelize folder which contains all the generic sequelize code, so if someone wants to add any other dbs supported by that there shouldn't be much code required. In fact, if we change to using JWTs all you should need is a new ```sequelize_config.js``` and an ```index.js``` to export all the variables.

server/db/sequelize/models/users.js
```diff
@@ -0,0 +1,95 @@
+import Promise from 'bluebird';
+import bcryptNode from 'bcrypt-nodejs';
+const bcrypt = Promise.promisifyAll(bcryptNode);
```
choonkending: @mjewell Do you see value in adding `bluebird` as an added dependency as opposed to using `es6-promise`?

On one hand, it promisifies every function, but just being careful about adding another promise library (although bluebird appears to be faster hehe)

docs/databases.md
```diff
@@ -33,3 +39,34 @@ You will need to add a folder after `/db` with [may] contain the following ORM s
...
+
+**Run your Postgres server**
+```bash
+postgres -D /usr/local/var/postgres
+```
+
+**Setup your postgres database**
+```bash
+createdb react_webpack_node_development
+sequelize db:migrate
```
choonkending: For local development, it might be worth noting that we need to run
```bash
createuser root
```
to create the role `root` that our default settings has.

choonkending: Given that we install sequelize without the `-g` flag, we could either run:
```bash
// in this project directory
node_modules/.bin/sequelize db:migrate
```
or
add a `sequelize db:migrate` to `package.json` and use `npm` to run the command?

server/db/postgres/sequelize_config.js
```diff
@@ -0,0 +1,19 @@
...
+  test: {
+    username: process.env.PGUSER || 'root',
+    password: null,
+    database: 'react_webpack_node_test',
+    host: '127.0.0.1',
+    dialect: 'postgres'
+  },
+  production: {
+    use_env_variable: 'POSTGRES_DB_URL'
+  }
```
choonkending: Do we need to specify the dialect as well as other config here?

Otherwise we could add some docs on how to test production locally on your machine

server/db/sequelize/models/users.js
mjewell: For this I was mostly concerned about not handling all the cases properly, where I know bluebird is safe. I'm not sure es6-promise has a promisify method, but if you look at [bluebird's promise anti-patterns](https://github.com/petkaantonov/bluebird/wiki/Promise-anti-patterns) they mention:

> The generic promisification is likely to be faster because it can use internals directly but also handles edge cases like libraryFunction throwing synchronously or using multiple success values.

I know I could handle these cases myself but there could be more I'm not aware of on top of this.

docs/databases.md
mjewell: Yes, you can do that. Locally I was just changing the url in constants and the username in sequelize_config but maybe this is better, or at least documenting that. Which do you prefer?

Good point for sequelize. Will update to include an npm script

server/db/postgres/sequelize_config.js
mjewell: It looks like sequelize figures out the dialect from the url, since that is of the form:

```
dialect://asdf:ghjk@whatever:1234/qwerty
```

You can see it pulls out the 'postgres' part [here](https://github.com/sequelize/sequelize/blob/master/lib/sequelize.js#L132)

mjewell: I will add those config settings to the production block too since sequelize preferentially uses the use_env_variable option, that way you can just comment that single line to test prod locally

mjewell: Now it uses the env variable in prod but falls back to similar settings to dev, so as long as you don't have the env variable set locally to something you didn't want and you have the prod database created, everything will work just like it did in dev

choonkending:
@mjewell Thanks very much for all the hard work you put into this PR!

Tested and working (dev and prod).

PR #194, opened 2016-04-18
Fix redux devtools

mjewell:


app/store/configureStore.js
```diff
@@ -12,19 +12,17 @@ import createLogger from 'redux-logger';
...
-    middleware.push(reactRouterReduxMiddleware);
+    middleware.push(createLogger());
   }
 
-  const finalCreateStore = applyMiddleware(...middleware)(createStore);
-
-  const store = finalCreateStore(rootReducer, initialState);
+  const store = createStore(rootReducer, initialState, compose(
+    applyMiddleware(...middleware),
+    typeof window === 'object' && typeof window.devToolsExtension !== 'undefined' ? window.devToolsExtension() : f => f
```
choonkending: Should this be added behind a __DEV__ flag?

mjewell: I guess it depends on what you want. According to the repo, one of the advantages of it is:

> Having DevTools even in production without any drawbacks.

It's only problematic if the end user has redux devtools installed, and even then only if you care about them seeing the internals of your JS, which shouldn't really have any secrets in it anyway. I still kinda see the hesitation, but personally I think it's fine. Let me know what you think and I will change it accordingly

choonkending: > It's only problematic if the end user has redux devtools installed, and even then only if you care about them seeing the internals of your JS, which shouldn't really have any secrets in it anyway.

Agreed.

Let's go with having it in production. If there are issues with it, I think we will find out really quickly.

#justdoit

choonkending:
@mjewell Thanks for this. Merged! Will bump up version soon!

PR #201, opened 2016-04-21
fix duplicate fetch in WrappedComponent.need

### LINKED ISSUE #166: fetchTopics called twice on server-side render
Hi,

First of all thanks for your boilerplate. I'm just starting out and having all the pieces together helps me out a lot!
While trying to figure out which piece is doing what, I noticed that on the initial page render fetchItems is called twice.
I think this is caused by the line

const needs = components.reduce( (prev, current) => {
    return (current.need || [])
      .concat((current.WrappedComponent ? current.WrappedComponent.need : []) || [])
      .concat(prev);
    }, []);

in api/fetchComponentDataBeforeRender.js

where the wrappedComponent line causes the 'need' to be added to the 'needs', leading to fetchItems being called twice. Removing the line with wrappedComponent solves this problem, but I'm not sure if this is somehow caused by the react-router setup or if there's a better solution..

AJIh:
fix #166.

In ES6, `static` properties are the class' own properties. So

```javascript
class Vote extends Component {

  static need = [  // eslint-disable-line
    fetchTopics
  ];
 ...
}
```
is transpiled into this by babel

```javascript
function Vote() {}
...
Vote.need = [fetchTopics];
```

And in `react-redux` source code, `connect` returns `hoistStatics(Connect, WrappedComponent)`, which merges WrappedComponent's own properties into Connect. So `current.need` is identical with `current.WrappedComponent.need`.

### DIFF
--- a/app/api/fetchComponentDataBeforeRender.js
+++ b/app/api/fetchComponentDataBeforeRender.js
@@ -7,7 +7,6 @@
 export function fetchComponentDataBeforeRender(dispatch, components, params) {
   const needs = components.reduce((prev, current) => {
     return (current.need || [])
-      .concat((current.WrappedComponent ? current.WrappedComponent.need : []) || [])
       .concat(prev);
     }, []);
     const promises = needs.map(need => dispatch(need(params)));


choonkending:
Sorry took me a while to get up to speed with this. Thanks for fixing this issue!

**References**
- [hoistStatics in react-redux](https://github.com/reactjs/react-redux/blob/master/src/components/connect.js#L365)
- [hoistStatics Lib](https://github.com/mridgway/hoist-non-react-statics/blob/master/index.js#L27)

CntChen:
I write my code use `ES6` and face duplicate fetch problem too.
 After testing:
* `react-redux@4.4.5.connect` method wrap React Component's static properties into new Component;
* `react-redux@2.1.2.connect` don't.

so I think this issue is not relevant to `ES6`, but the version of `react-redux`.

PR #203, opened 2016-04-22
Fetch component data when frontend routing

### LINKED ISSUE #202: Front end routing does not fetch data for component
`Vote.need` functions are only called in server side. When clicking `Ninja Ocean` to navigate to index route, the `Vote` component's props are not fetched at front end.

AJIh:
fix #202 

When navigating from `/login` to `/`, Vote component data is not fetched.

Add a callback function when doing routing to fetch required data.

### DIFF
--- a/app/client.jsx
+++ b/app/client.jsx
@@ -5,6 +5,7 @@ import { Router, browserHistory } from 'react-router';
 import { syncHistoryWithStore } from 'react-router-redux';
 import createRoutes from 'routes.jsx';
 import configureStore from 'store/configureStore';
+import { fetchComponentDataBeforeRender } from 'api/fetchComponentDataBeforeRender';
 
 // Grab the state from a global injected into
 // server-generated HTML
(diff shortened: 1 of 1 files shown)


app/client.jsx
```diff
@@ -14,11 +15,25 @@ const store = configureStore(initialState, browserHistory);
 const history = syncHistoryWithStore(browserHistory, store);
 const routes = createRoutes(store);
 
+/**
+ * Callback function handling frontend route changes.
+ */
+function onUpdate() {
+  // Prevent duplicate fetches when first loaded.
+  if (window.__INITIAL_STATE__ !== null) {
+    window.__INITIAL_STATE__ = null;
```
choonkending: on second thought, why do we have to reset the `__INITIAL_STATE__` to null if we're returning? Not sure if this line is needed?

AJIh: In my opinion, `__INITIAL_STATE__` is used to create initial store and match the markups server rendered. So if you navigate along these paths:

`/`(server rendered): `onUpdate` is triggered first time. `__INITIAL_STATE__` contains all required data in this page. No need to refetch.
↓
`/dashboard`: `onUpdate` triggered second time. `__INITIAL_STATE__` doesn't contain required data in this page. Call `fetchComponentDataBeforeRender`.
↓
`/`: `onUpdate` triggered again.  `__INITIAL_STATE__` is outdated for this page. Call `fetchComponentDataBeforeRender`.

To summarize, after loading server rendered page,  `__INITIAL_STATE__` is useless. When navigating to other pages  `__INITIAL_STATE__` doesn't contain required data. When navigating back to the first page `__INITIAL_STATE__` is outdated and needs refetch.

choonkending: Thanks for that, I get what you mean now.

We only need `window.__INITITAL_STATE__` for configuring the store on server-rendered pages.

The check is only for preventing duplicate fetches when `server-render` -> client side kicks in and doesn't need to refetch. 

Every subsequent client `onUpdate` call should then trigger a `fetchComponentDataBeforeRender` because `__INITIAL__STATE__` is outdated. 👍

choonkending: Really good explanation btw. Good work! I think this is the best approach we can take for this.

choonkending:
Also added a bit of our discussion https://github.com/choonkending/react-webpack-node/commit/b1363846fd642b3b6f5888343fd86a98f271afe5#diff-18882a44c2913bd42f984881e8e0deb5R24 into `client.jsx` in case people get a bit confused (like I did). 

Thanks!

PR #206, opened 2016-04-25
move login/register into a form

huyphamily:
It felt weird not having the expected behavior of enter to work on the login/register screen. I wrapped the inputs in a form and added an onSubmit handler. This way, both the actions of clicking the button and pressing enter will submit the post to login/register.

### DIFF
--- a/app/containers/LoginOrRegister.jsx
+++ b/app/containers/LoginOrRegister.jsx
@@ -17,8 +17,7 @@ class LoginOrRegister extends Component {
   constructor(props) {
     super(props);
     this.toggleMode = this.toggleMode.bind(this);
-    this.onLoginSubmit = this.onLoginSubmit.bind(this);
-    this.onRegisterSubmit = this.onRegisterSubmit.bind(this);
+    this.handleOnSubmit = this.handleOnSubmit.bind(this);
   }
 
   onLoginSubmit() {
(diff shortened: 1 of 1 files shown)


app/containers/LoginOrRegister.jsx
```diff
@@ -17,28 +17,27 @@ class LoginOrRegister extends Component {
...
-      password
-    }));
+
+    if (isLogin) {
+      dispatch(manualLogin({
+        email,
+        password
+      }));
+    } else {
+      dispatch(signUp({
```
18601673727: Not every login request is a register request, so maybe you should do more logic check before dispatching it.

huyphamily: I'm keeping the logic the same as the boilerplate. The isLogin property on the state determines whether or not the user is trying to login or register. When it is true, the user is trying to login. When it is false, the user is trying to register. Take a look at the old renderButton method on this container, the logic will look similar. The action creator toggleLoginMode toggles between the two mode, and there is only two mode in this case.

choonkending:
Thanks for contributing!!

I agree that enter not working is an unexpected behavior. Will review and merge hopefully late today!

choonkending:
Thanks again for contributing!

huyphamily:
Not a problem, happy to contribute. You have a very nice boilerplate!

PR #210, opened 2016-04-26
sequelize_config.js is safe to check in now

### LINKED ISSUE #209: should we still not commit sequelize_config.js?
https://github.com/choonkending/react-webpack-node/blob/master/server/db/sequelize/sequelize_config.js

The warning at the top was added back when the file was automatically generated, but it seems pretty generic and nice now. If we don't commit it, and try to build somewhere new, won't that be a pain? Is there any downside to committing it?

WestleyArgentum:
fixes #209


mjewell:
Thanks for cleaning this up

WestleyArgentum:
not at all, thanks for being so quick to respond!

choonkending:
Thanks for this! Good find! :+1

Issue #119, opened 2016-01-09
Add helmetJS


choonkending:
https://github.com/helmetjs/helmet

Better security. 
Also follow the steps here. We want good practices here.

https://strongloop.com/strongblog/best-practices-for-express-in-production-part-one-security/

psimyn:
I think all done @choonkending ?

choonkending:
not sure if I did it :(

choonkending:
Yep, have not done this yet. This is the helmet for security.

PR #214, opened 2016-05-01
Updated react-helmet to 3.1.0 and imporved code readability

18601673727:


choonkending:
Thanks for contributing! Do you mind making splitting this commit into upgrading react-helmet and the refactor into another commit?

I'd be happy to help out, but it would be good to have these changes in logical commits for other users to follow.

choonkending:
Thanks for taking the time to update the pull request.

I like (and agree) the changes:
- Renaming `api` to `middlewares`
- Upgrading react-helmet
- Fixed indentation
- Better and reduced code 👍 

However, some of the changes might need a bit of discussion.

- I like having `fetchComponentDataBeforeRender` because it is more explicit than `preRenderMiddleware` which might not give as much context? E.g. developers might read `preRenderMiddleware` might not immediately understand what the role of this middleware is unless they look at the code itself. Again, this is [one the harder things in computer science](http://martinfowler.com/bliki/TwoHardThings.html) lol. We could perhaps rename it to **`fetchComponentDataMiddleware`** if you think that's a better alternative.
- I prefer using the original signature of  `match` as opposed to 
```javascript
match({routes, location: req.url}, (err, redirect, props) => 
```
mainly because it is consistent with what react-router has named them [here](https://github.com/reactjs/react-router/blob/master/docs/API.md#match-routes-location-history-options--cb)

Please let me know if you disagree or perhaps we could discuss better module names.

choonkending:
I'd like to get the react-helmet commit merged but still have reservations regarding the naming.

Perhaps it's worth rebasing with upstream and creating a separate PR for only the following files?

```
// Add main repo as upstream
git remote add upstream git@github.com:choonkending/react-webpack-node.git
// Pull latest changes from master but play current commits after
git pull -r upstream master

git reset HEAD~2 (if the helmet commit (b0aa8152151b39e151bdffdecdd02a9d535eeed5) was the 2nd last commit)

// git add only the package.json + Helmet changes as a single commit. 

```

18601673727:
😊

Hi @choonkending,

Thanks for your kindly appreciation and offering this amazing framework!

- I can't agree more with the 2 hard things in CS and `fetchComponentDataMiddleware` is a pretty good name but what if we had a `need` list like this?
```es6
static need = [
  keepAlive,
  fetchTopics,
  computeSomethingBig,
  thirdpartyStatistic,
  syncOfflineData
]
```
Btw, I also thought regarding all the api callings as endpoint objects method calling is better than raw `fetch` style function calling, and that's why I hope #184 to been got merged soon.
- `match()` parameter changes are took from [here](https://github.com/reactjs/react-router-tutorial/tree/master/lessons/13-server-rendering#server-rendering), it's a little bit cleaner I thought and we 've already have a long comment above the `match()` function so no need to emphasize it too hard.

Also thanks for the git command suggestion, I'll try it tomorrow and make that done asap. = ]

choonkending:
@18601673727 
> 
```javascript
static need = [
  keepAlive,
  fetchTopics,
  computeSomethingBig,
  thirdpartyStatistic,
  syncOfflineData
]
```



Very valid point! If there are things that **do not involve fetching** but needs to be done before rendering, `prerenderMiddleware` is an excellent name. Thanks for pointing that out.

> we 've already have a long comment above the match() function so no need to emphasize it too hard.

Cool another valid point haha.

As for #184 I'll have a look again now. It would be good to have an extra 👀 on it as well.

18601673727:
Hi @choonkending, the commit has been splited now, I hope you like it! And I always pleasure to talk about naming if you really don't like it, there must be a perfect name for that procedure. 😄

choonkending:
Thanks buddy! The code looks good!

I'll pull and test locally before merging after work :)

choonkending:
@18601673727 I think it might have been a merge conflict resolution that might have missed out https://github.com/choonkending/react-webpack-node/pull/214/commits/ed373df210b5f264d2899daf2ee45c815f2b0f19#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R106. Could you add that package bump back in again?

choonkending:
Apart from that, I'm happy to merge (I've tested by bumping the version locally)!

18601673727:
@choonkending, It's not a conflict cause the current master already had that line modified which is the same as this PR, so there's no change been made by me. = ]

choonkending:
Sorry my bad! Derp moment here.

For some reason, I got `react-helmet` mixed up with `helmet` [here](https://github.com/choonkending/react-webpack-node/blob/master/package.json#L83). We don't even use it yet! I had been wanting to use it though. I must have accidentally introduced the dependency during https://github.com/choonkending/react-webpack-node/issues/119

choonkending:
Thanks for helping out @18601673727 !

Issue #220, opened 2016-05-06
Code refactors


choonkending:
**Description**
*JS*
- Use stateless functions if possible
- Ensure propTypes used everywhere
- use `mapDispatchToProps` in our connect functions.

**Current implementation**
```javascript
import {connect} from 'react-redux';
class A extends React.Component {
   constructor(props) {
     super(props);
     this.doSomething = this.doSomething.bind(this);
   }
   doSomething() {
          this.props.dispatch(someAction());
   }
   render() {
        return (<SomeComponent onClick={this.doSomething} />
    }
}

export default connect()(A);
```

**Moving forward**
```javascript
class A extends React.Component {
     render() {
        return (<SomeComponent onClick={this.props.onAction} />
     }
}

function mapDispatchToProps(dispatch) {
     return {
         onAction: () => dispatch(someAction)
     };
}

export default connect(null, mapDispatchToProps)(A);
```

*CSS*
- Vertical rhythm to CSS

huyphamily:
What are your thoughts on the short-hand version of mapDispatchToProps, especially when we are not renaming the action creator?

Here's an example:
```
import { createTopic } from 'actions/topics';

class A extends React.Component {
     render() {
        return (<SomeComponent onClick={this.props.createTopic()} />
     }
}

export default connect(null,{ createTopic })(A);
```

choonkending:
Thanks for the suggestion!

Yep I like it, unless we find the need to remap our action type creators. Although it would be less understandable. We can roll with the short-hand version and see how we like it.

```
import { createTopic } from 'actions/topics';

class A extends React.Component {
     render() {
        return (<SomeComponent onClick={this.props.createTopic} />
     }
}

export default connect(null,{ createTopic })(A);
```

I don't think we need to wrap the onClick handler in a function do we? We can just pass in the `prop` as is.

huyphamily:
I agree, you don't need the wrap in this case. As I understand it, and correct me if I'm wrong, if we want to pass in some parameters, then we need to wrap it or use .bind.

`return (<SomeComponent onClick={() => this.props.createTopic('hello')} />`
`return (<SomeComponent onClick={this.props.createTopic.bind(this, 'hello')} />`

choonkending:
@huyphamily You would need to bind it if you're using a function within that component (not the prop) I think. 

Also, using an anonymous function in the `render` method is not a good idea because you would not be able to use Pure Render Mixin or optimise rendering.

PR #221, opened 2016-05-09
So long redux-thunk, hello redux-axios!

18601673727:
This PR aims to provide the shinning `HTTP Client` feature to our beloved framework. 🎁

Current progress:
- [x] Replace `redux-thunk` with <del>`redux-axios-middleware`</del> `redux-axios`
- [x] Refactor of `Topic` components
- [x] Refactor of `User` components
- [x] Refactor of `Message` components, also resolve https://github.com/choonkending/react-webpack-node/issues/230 and https://github.com/choonkending/react-webpack-node/issues/229
- [x] Promise Support, waiting for https://github.com/svrcekmichal/redux-axios-middleware/pull/9
- [x] Multiple HTTP Client support, waiting for https://github.com/svrcekmichal/redux-axios-middleware/issues/6#issuecomment-219218914
- [x] Global loading bar, <del>waiting for https://github.com/mironov/react-redux-loading-bar/issues/2</del>, please try with `Chrome Dev Tool -> Network -> Throttle GPRS`
- [ ] Interceptors, the foundation of OAuth support(maybe cover https://github.com/choonkending/react-webpack-node/pull/184), currently test wanted.


choonkending:
@18601673727 Thanks for contributing - you're on fire! 🔥 

The code changes I can see are **really good**. However, it might be worth discussing the advantages of using `redux-axios-middleware` over `redux-thunk`, as well as the other bullet points in this PR so that we have clear visibility of your thought process - I'm sure they are valid points!

The reason why I'm a bit hesitant with making dependency changes are because some of the existing packages/libraries are used by users within this repo already, and it would make updating existing projects if we were to swap that dependency. 

So although it might be a bit time consuming (and frustrating I bet) to wait for feedback, we could split each of these into new issues to discuss. My concern here is that I **do not want your effort to go to waste** if there e.g. we decide **for/against** using a particular framework/pattern/library.

It would be really helpful for me/and other maintainers to understand the problem and solution - so that we can get changes merged in quicker!

Here's an example of an issue:

---------------------------------
E.g.
Example issue

**Problem**
Redux-thunk does not handle promises well

**Proposed Solution**
Use `redux-axios-middleware`

**Pros**
- It has an awesome HTTP client baked into the middleware!

**Cons**
- It might tie us to `axios` and limit the flexibility of using other HTTP libraries that our users use. (Making it very coupled to `axios`).

**Alternatives**
- `<Fill in your libraries>`

----------------------------

**I'm guilty of not documenting my problems/pros/cons in this repo as well**, but that's what I'm hoping to improve on to give visibility and spark better discussions among the community as well. I will come up with an issue, PR template so that it would make it easier for you to communicate your concerns with this framework.

Ideally we have `1 issue <-> 1 PR` so that it's easier to review. Code refactors can be done as separate PRs as well.

I hope you don't mind this long post! My main concern is that I wouldn't want hard work to not get merged in, so preferably we could go with a more `discussion -> code` approach (unless it is a quick code snippet).

Thanks for your hard work!

choonkending:
I'll go through the initial issue again and possibly document the problems - it might just be me losing context of the issues I'm dealing with 🔨

18601673727:
@choonkending, Yeah I directly pulled this merge request without opening an issue or informing you before cause I planned to start a project base on this framework and use in my team, but I found out the `thunk` way of doing things doesn't very fit my practice so much and things are going complicated with it. So I decide to do it and see how the community react, I know there are lots of people already wrote lots of code by the current pattern, they have more right to speak than I do. 😃 

On the other hand, I don't have much theoretical basis on doing this, and even don't know how far I can go, but I'll give it shot and see whether the pros outweigh the cons.

choonkending:
>  I'm in a hurry to start a project base on this framework and collaborate with my team,

I'm excited you're using this boilerplate for your project!

> but I found out the thunk way of doing things doesn't fit my practice so much and things are going complicated with it.

Can you describe what things don't fit in your practice? Is it related to #225? 

> I don't have much theoretical basis on doing this, and even don't know how far I can go, but I'll give it shot and see whether pros outweigh the cons.

Not a problem! We're all learning! :)

18601673727:
Yeah, #184 and #225 are definitely related but not enough, **I'm not intending to start a fight** with `thunk` fans, but I don't think  the best way of dealing async actions(Lazy Evaluation) is `thunk`:

1. **It doesn't provide a "client"**
Maybe you'll have lot's of api groups to wrap, one for google, one for github, one for your backend, how to organize them in action creators? By writing “action creator creators”? 

2. **It will make your code dirty**
[Just like your comments here](https://github.com/choonkending/react-webpack-node/blob/master/app/actions/topics.js#L77-L80), imagine if you're on a large project, how many action creators you have to make, and how hard to keep them in order or reusing? Api is just an abstraction of data, let's say it's just a remote model aggressively.

3. **It's a difficult-to-learn solution.**
If you want to really understand `thunk` you must have some experience of functional programming like `Haskell`, `Lisp` etc, This is pretty good but is a huge cost to my team. 

So I prefer to use another approach like `promise`(monad) which has beat `thunk` to been a spec of ES6. I don't know who brought `Axios` here, but it seems good so I add `redux-axios-middleware` in, but it seems far from stable right now, I'm very new to React Community, so I guess there must be a mature way to achieve what I'm trying to.

Now I think I need to do some research then catch up you later. = )

18601673727:
Found several alternatives:
- https://github.com/pburtchaell/redux-promise-middleware
- https://github.com/yelouafi/redux-saga
- https://github.com/raisemarketplace/redux-loop
@choonkending

choonkending:
@18601673727 A bit backed up past few days, I'll have a look at the alternatives and your points slowly when I get the chance!

18601673727:
Hi @choonkending, hows your weekend, I bring some good news here, one is the author of `redux-axios-middleware` has updated his work and I applied it to this PR so "Promise Support" is done. Another thing is the refactors of left components which related has been done! Please take a look and better try them on your local side. 😄

choonkending:
Sure!! Thanks for contributing! I've been a bit behind in reviewing PRs and
issues last week. I'll put in more time this week!

Good work!

choonkending:
> It doesn't provide a "client"
I still like the concept of separating the asynchronous action from whatever implementation you select. I don't really understand `how to organize them in action creators`. Isn't it a matter of writing the action creators in a way that allows you reuse regardless of Github, Google, etc?

> It will make your code dirty
Agreed

>  This is pretty good but is a huge cost to my team.
I understand the cost of learning, but on a separate note I believe understanding more functional programming helps in the long run, because you will find yourself writing shorter, concise code that is less likely to break over time. :P
 
I still need to do a bit of reading, so far this is what I've come across:
- [redux-effects](https://github.com/redux-effects/redux-effects)
- [Good discussion here](https://github.com/reactjs/redux/issues/1528#issue-141756305)

I'm also going through the alternatives you suggested.There's still a lot I do not know about promise middleware, reducer composition and side effects in general and I would like to read up more on it before selecting a good solution. 

Given you've made some changes unrelated to redux-thunk/promise middleware, e.g. the messages commit.  Perhaps it's worth creating an additional PR with just the messages refactor if possible? Keeping the PR small makes it easier for me to review and merge.

18601673727:
>Given you've made some changes unrelated to redux-thunk/promise middleware, e.g. the messages commit. Perhaps it's worth creating an additional PR with just the messages refactor if possible? Keeping the PR small makes it easier for me to review and merge.

I'm afraid if separate message refactor out might cause these file conflicts while merging this PR in the future:
```
app/containers/LoginOrRegister.jsx
app/containers/Navigation.jsx
app/actions/users.js
```
> I still like the concept of separating the asynchronous action from whatever implementation you select. I don't really understand how to organize them in action creators. Isn't it a matter of writing the action creators in a way that allows you reuse regardless of Github, Google, etc?

In fact I mean there need to be a layer to abstract different endpoints, and then in the network related actions, everything you left to do is specify which endpoint this request should pick. So you don't have to configure or initial them in the action functions(too bad) or by hand(even worse), it maximize the concise of actions code, doesn't it?

> I understand the cost of learning, but on a separate note I believe understanding more functional programming helps in the long run, because you will find yourself writing shorter, concise code that is less likely to break over time. :P

Can you tell me what makes this PR less FP compare to the master branch? :laughing: Actually `redux-axios-middleware` did a great job and it's enough for single endpoint situation now.

choonkending:
Hahaha the Changelog is going to be *huge* when we merge this in.

Looking at the code, I agree that a lot of boilerplate code has been removed from when using `redux-thunk`.

--------------------------------------------------------------------------
Just to get a common understanding of the problems we wish to solve:
### Handling Asynchronous actions with Redux
Current implementation:
- Redux-Thunk
- `api/promiseMiddleware`

*Issue with the current implementation*
- A lot of code
  - I noticed in this PR using redux-axios-middleware, specifying a `payload` and `types` greatly simplifies things by using 1 function instead of 3 functions.
- Promise support

### Authentication for Server-Side rendering
> if user comes to website with active session cookie and you call server side rendering through static needs methods, you don't see user session information, because the request is made from server to server

- Related Issue: #184
- Related PR: #225
There has been good work done from @18601673727 and @Epochi regarding this complex issue.

Even though this one is in the works, I think it's safe to say it relies on using interceptors to attach or remove cookies using `axios`. 

------------------------------------------------------------------------------------------------------

There are still a couple of points I'd like to make:
- `redux-thunk` helps with asynchronous actions
- `promiseMiddleware` (which can be replaced with [redux-promise](https://github.com/acdlite/redux-promise) or [redux-promise-middleware](https://github.com/pburtchaell/redux-promise-middleware) if we like) allows the usage of promises
- `redux-axios-middleware` is an excellent middleware that handles data fetching with axios as a HTTP client

Asynchronous action are not necessarily always Ajax calls. 

From the docs of redux-axios-middleware, it shows that the chaining of the promises occurs from within the component or wherever the `dispatch` function lives. (I might have totally misread or misunderstood the docs for redux-axios-middleware), but I would much rather have any side-effects or asynchronous actions done *within* the action creator, and not within the component, mainly because of preference. Not saying that doing so is bad. 

- I do not see why we cannot use `thunk` and `promises` together. 
**Thunk** middleware is simple and excellent because it gives the user an idea of *what* happens when the function is executed. It handles all the `dispatch`-ing of actions from the actionCreator itself, rather than abstracting that to a middleware. It helps the code to seem less *magic*.

I agree that there's a bit too much repetitive code within the action Creators. So perhaps we could modify our promise middleware to also accept `types` for request, success, failure.

-----------------------------------------------------------------------------

The **tl;dr** version is that I agree with the problems you've listed, but still feel unsure about `redux-axios-middleware` as a dependency + middleware as it makes it a big too magical, **unless** we find that there is no other way but to intercept the axios instance for authentication (in #184). I think we can address the problems you wish to solve by either:
- Changing a bit of our promiseMiddleware
- Introducing an existing promise middleware that is more generic and not implementation specific

Given that this is a big PR, I might have a lot of comments, so bear with me 😆 

Sorry for the extremely wrong write up. I took a long time to come up with the above points. 

I want to do a quick code refactor to address #220 so it might be worth me doing that really quickly so you know how I plan the direction of the structure of the components.

choonkending:
Also a good read by Dan Abramov http://stackoverflow.com/questions/35411423/how-to-dispatch-a-redux-action-with-a-timeout/35415559#35415559

18601673727:
@choonkending, It takes me too much time to go through all your comments, I thought about it for few days and this's why I reply you so late (sorry, but I hope you don't mind 😄 ) 

Dan Abramov had made a very good point about `react-thunk`, like:
> For simple apps, the approach should suffice.

or:

> For some apps, you may find yourself in a situation where your asynchronous control flow requirements are too complex to be expressed with thunks. For example, retrying failed requests, reauthorization flow with tokens, or a step-by-step onboarding can be too verbose and error-prone when written this way.

So, here's the deal, if you wanna to expand the possibility of building larger scale applications, you may not want to use `react-thunk` because what its author said. And in my opinion, all the objection between you and me is **how to deal with the component communication problem** -- should I written that part of code in actions or in components?

Well, my answer is **components** because I believe in actions and reducers both should been pure functions, they doesn't have to bring **side-effects**, they only should care about one thing -- **data**.

`react-saga` gives another answer -- adding an extra layer of handling components' communication(I don't quite sure about this). It seems like a bigger picture than `react-axios-middleware`, but after [saw this project](https://github.com/mxstbr/react-boilerplate) makes me feel more wired to use it in the future.

`react-promise-middleware` seems to been the most comparable alternative, it has the most similar logic to solve the problem, it's more mature in the community, it compatible with `react-thunk` out-of-the-box, **but wait** what did just I say? It doesn't close the door. `react-promise-middleware` is too more focus on `Promise` which means too raw for the larger projects, and this lead out **they'll not have HTTP Client feature** for now and forever.

caranicas:
So is `react-axios-middleware` the direction we are headed? I just want to know what to get out in front of, I was only just starting to really use thunks, and was getting more interested in `react-saga` as a way building out flows. from everything I had read it seemed like the next logical step and a good excuse to start familiarizing myself with generators. I was about to post an issue bringing up the idea of moving over to them but then I read this post, and it seems like a lot of work has been done towards axios then I am very happy to move in that direction. It seems like the solve the same issues.

choonkending:
Sorry for getting back to you guys so late. 

> So is react-axios-middleware the direction we are headed?

I'm still sitting on the fence on this one, as I am still not liking the hard dependency on `axios` as a middleware. It's too magical. 

@18601673727 and I both **agree** on the fact that our action creators are getting too verbose and large.

> Well, my answer is components because I believe in actions and reducers both should been pure functions, they doesn't have to bring side-effects, they only should care about one thing -- data.

If you mean where the request gets made, doesn't `redux-axios-middleware` make the call in the middleware?

> this lead out they'll not have HTTP Client feature for now and forever.
I don't think this is a bad thing. I'd like to keep this boilerplate as agnostic as possible. 

Perhaps I'm getting lost in the code as this PR is tackling several issues at once, and I'm finding it hard to understand or get enough context.

@18601673727 Sorry about this, but can we split this out?
- loading bar
- moving files to different folders
- middleware
- etc

These changes can be merged in with separate reviews/discussions, but it will help with chipping bits off this giant PR.

@psimyn @caranicas @Epochi would like to know your thoughts as well.

caranicas:
I've gone through the docs for `axios`, the `redux-axios-middleware`, and `redux-saga`, trying to get a sense of each of them on what they bring to the table. I'm going to try to synthesize that, so we can have a common place to start this discussion from

As I understand it both of the middlewares solve the issue of multistep actions, but they have different focuses. `redux-saga` uses generators, and  is built to handle multi step processes.  `redux-axios-middleware` seems to be focused on server responses, and using promises to chain actions from those responses. 

Personally I prefer the declarative nature of `redux-saga`, being able to compose sagas that represent flows sounds like a very nice way to work. I built some user on boarding features into my application two weeks ago, and If I had being using `redux-saga` my code would be a lot cleaner than it is right now. 

What I am most unclear about, is whether or not you can use `redux-axios-middleware` to chain non server related actions. If it can't then I think that `redux-saga` is the clear choice, if it can, and it is just a matter of building flows with promises vs generators then I think there is more to discuss about pros and cons of promises vs generators.

app/actions/network.js
```diff
@@ -0,0 +1,29 @@
+import { NETWORK_EXCEPTION } from '../types/message';
+
+export function fetch(type, data) {
```
psimyn: can these get named something different (possibly use the HTTP verbs)? Calling `fetch` with a different signature to `window.fetch` could lead to weird mistakes if you forget to import

app/components/TopicItem.jsx
```diff
@@ -46,9 +46,7 @@ export default class TopicItem extends Component {
 }
 
 TopicItem.propTypes = {
-  text: PropTypes.string.isRequired,
-  id: PropTypes.string.isRequired,
-  index: PropTypes.number.isRequired,
+  topic: PropTypes.object.isRequired,
```
psimyn: any reason for changing this other than LoC? passing in `id` and `text` makes props comparisons simpler (and indicate what is actually used).

app/containers/Message.jsx
```diff
@@ -6,16 +6,15 @@ import styles from 'css/components/message';
...
+const Message = ({style, text, dismissMessage}) => (
+    <div className={cx('message', [style, {show: text && text.length > 0}])}
+      onClick={dismissMessage}
+    >{text}</div>
 );
 
 Message.propTypes = {
-  message: PropTypes.string,
-  type: PropTypes.string,
+  style: PropTypes.string,
```
psimyn: why the change for these prop names? `style` in particular is usually used for an object (for inline styles)

psimyn:
@18601673727 I left a few small comments - I'm of similar mind to @choonkending that it'd be easier to review this if split out into sections (and some of the renaming). 

Can you please split out the different parts of this (and into separate branches instead of master) - it will be easier to add the parts.

Overall, great work reducing the boilerplate! I'm happy to proceed with this largely as-is, then we can swap between sagas/axios-middleware later. Boilerplate repos are a good place for experiments migrating between different techniques :)


@caranicas:

> What I am most unclear about, is whether or not you can use redux-axios-middleware to chain non server related actions.

This was my first thought as well - from what I understand it is only intended to handle HTTP requests (cursory skim - may be completely wrong). I think your explanation of the differences is pretty spot on, especially 'discuss about pros and cons of promises vs generators [vs libraries]'. Ultimately I think the choice shouldn't matter hugely - can always swap it later :)

caranicas:
@psimyn  The only thing that made me think that the `redux-axios-middleware` might be able to work for all types of actions was the code block in the docs
```
"Another example might be a set of actions that you want dispatched together."

Promise.all([
  dispatch(foo()),
  dispatch(bar()),
  dispatch(bam()),
  dispatch(boom())
 
]).then(() => {
  dispatch(
    loginSuccess(
      token
    )
  )
})
```
But this is a bit ambiguous, it is unclear if  `foo`, `bar`, `bam` and `boom` get dispatched sequentially or all at the same time.

psimyn:
`Promise.all` should run them in parallel, and the results all in the next `then` (in same order). This means they can't depend on each other; unsure how you'd handle that.

So I think you'd need a different interface for non-HTTP actions - which I think is an advantage of redux-saga way

app/components/TopicItem.jsx
18601673727: Yeah, you're right, will change this later, thank you!

app/containers/Message.jsx
18601673727: Ahh, it's because `type` is so widely used in this project, type everywhere seems like a very confusing plan. Other suggestions would been kindly welcome.

app/actions/network.js
18601673727: @psimyn 
When you call fetch() without import it, here's what you gonna get:
```
ReferenceError: fetch is not defined
   at fetchTopics (/Users/xxx/public/assets/server.js:28867:11)
   at /Users/xxx/public/assets/server.js:32693:22
   at Array.map (native)
   at preRenderMiddleware (/Users/xxx/public/assets/server.js:32692:11)
   at /Users/xxx/public/assets/server.js:125:42
   at /Users/xxx/public/assets/server.js:24806:6
   at /Users/xxx/public/assets/server.js:22685:12
   at done (/Users/xxx/public/assets/server.js:23167:20)
   at /Users/xxx/public/assets/server.js:23173:8
   at getComponentsForRoute (/Users/xxx/public/assets/server.js:23357:6)
   at /Users/xxx/public/assets/server.js:23412:6
   at /Users/xxx/public/assets/server.js:23172:6
   at Array.forEach (native)
   at mapAsync (/Users/xxx/public/assets/server.js:23171:10)
   at getComponents (/Users/xxx/public/assets/server.js:23411:29)
   at finishEnterHooks (/Users/xxx/public/assets/server.js:22679:36)
   at runTransitionHooks (/Users/xxx/public/assets/server.js:23011:6)
   at runEnterHooks (/Users/xxx/public/assets/server.js:23054:11)
   at /Users/xxx/public/assets/server.js:22672:43
   at runTransitionHooks (/Users/xxx/public/assets/server.js:23011:6)
   at runChangeHooks (/Users/xxx/public/assets/server.js:23071:11)
   at finishMatch (/Users/xxx/public/assets/server.js:22669:42)
```
_Also note this or any "forgot import errors" will not automatically occurred, you'll only receive this after restart of the server, and I think that might be a another bug unrelated with this PR._

18601673727:
I think `redux-saga` is not so related to my goal here, it's an extra layer for doing async things and flow control, [seemed doesn't conflict with `redux-axios`](https://github.com/yelouafi/redux-saga/blob/master/examples/real-world/services/api.js). so I would leave it to people who want to do it so.


@caranicas 
Have you tried this?
```es6
Promise.all([
  dispatch(foo()),
  dispatch(bar())
])
.then(() => {
  dispatch(bam());
})
.then(() => {
  dispatch(boom());
})
.then(() => {
  dispatch(
    loginSuccess(
      token
    )
  )
})
```

app/containers/Message.jsx
psimyn: ah, excellent point - `type` kinda indicates data type or something. I have seen success/warning/error passed in as boolean props, though that can get regretably busy if there's lots of options. It does look tidy though...

```
<Message text="Good news!" success />
```

choonkending:
Perhaps we'll kick off a redux-saga branch/PR and see how we go with that. 

In the meantime, we can also incorporate your excellent changes with the code refactors (but possibly in another PR).

18601673727:
@choonkending Yeah, I will split this PR into several minor PRs in couple of days, But honestly I don't think the integration of redux-saga(yield/generator) is necessary, but that depends on how you treat this boilerplate as.

caranicas:
What goal does that axios middleware solve for you that saga middleware cannot?

18601673727:
@caranicas Something like https://github.com/mgonto/restangular or https://github.com/guzzle/guzzle.

18601673727:
@choonkending This PR has been splitted as your wish, let me know if there anything else I can help.

app/containers/Message.jsx
18601673727: How does `color` sounds to you?

choonkending:
I'm closing this and reviewing them individually! :)

Issue #171, opened 2016-03-13
fetchComponentDataBeforeRender doesn't get access to session


Epochi:
Hey,

If user is already logged in and coming to the site, I'm unable to get user session data when using this

app/Containers/Votes.jsx
```
class Vote extends Component {
....
  static need = [
    fetchTopics
  ]
...
}
```
to call 
server/config/routes.js

`app.get('/topic', topics.all);`

server/controllers/topics.js
```
exports.all = function(req, res) {
   console.log(req.user) //returns undefined
   console.log(req) //returns shows empty passport object
};

```

I think it's because of fetchComponentDataBeforeRender, it dispatches the action from itself and not from the component, so it doesn't pass the user. Because I can see that passport deserealizing the user before the promise is called  and these console logs

```
'app.get('/topic', fiunction(req,res){
console.log(req.protocol + '://' + req.get('host') + req.originalUrl); 
//returns http://localhost:8080/top
})
  app.get('*', function (req, res, next) {
   //  this fires before /topic as it should
   // req.user returns the user
   // console.log(req.protocol + '://' + req.get('host') + req.originalUrl); //returns https://www.myaddress.com/
    App.default(req, res);
  });
```

Is there any way to see the session object in this situation?

Ultimately my goal is to detect if the user is logged in and if he is, perform topics query and check which topics he has already voted on, to change the color of upvote arrow. Which I'm able to do if calling fetchTopics from componentsWillMount() inside of Vote.jsx

choonkending:
@Epochi Nice use case :+1: for checking the user.

> Ultimately my goal is to detect if the user is logged in and if he is
Don't we already do this [here](https://github.com/choonkending/react-webpack-node/blob/master/app/server.jsx#L63). You can access the redux store to check if the user is authenticated.

You could probably modify [this](https://github.com/choonkending/react-webpack-node/blob/master/app/actions/topics.js#L128) function to conditionally check if the user is already logged in?

Let me know if that helps

Epochi:
@choonkending  hey, thanks for the answer!

You are correct, both of those correctly get the user session, but when fetchTopics() gets called from through fetchComponentDataBeforeRender, the axios promise comes from the server itself.

So User sends GET from his ip with his session data,cookies, and then fetchComponentDataBeforeRender sends a new axios GET from localhost and creates new request object which doesn't have req.user or req.session. So I'm unable to identify if the request was made from the authorized user.

[...]

PR #225, opened 2016-05-09
authentication for server side render with axios interceptors.

Epochi:
**UPDATE 2016/11/30**
This _still_ method bleeds data between users
**IMPORTANT**

Same as #184  but with significantly lower footprint

Before:
if user comes to website with active session cookie and you call server side rendering through static needs methods, you don't see user session information, because the request is made from server to server.

Now:
when logged in user comes to website, we attach his cookie to all requests for the duration that preRenderMiddleware function renders all the promises, then we delete the interceptor.

Pros:
- All server-side rendering can be done as if user was authenticated
- We don't need to modify specific routes with express to get user specific data on server-side rendering, and we can just use static need methods that work for unauthenticated and authenticated users

Cons:
- Relies on axios
- If developer is using interceptors for something else, then they'd need to modify preRenderAuthentication.js because it deletes all other request interceptors on default axios instance

### DIFF
--- a/app/api/preRenderAuthentication.js
+++ b/app/api/preRenderAuthentication.js
@@ -1,0 +1,19 @@
+import axios from 'axios';
+
+/*
+If it's initialized with a cookie, it sets it as a global interceptor
+if it gets called with no arguemnts, it deletes *all* request intercepotrs
+*/
+
+export default function ssrAuth(cookie){
+    if(arguments.length === 0){
+        axios.interceptors.request.handlers = [];
+    }else{
+    axios.interceptors.request.use(function(config) {
+            config.headers['cookie'] = cookie;
+    }, function(error) {
+      return Promise.reject(error);
+        });
+    }
+}
+
--- a/app/server.jsx
+++ b/app/server.jsx
@@ -7,6 +7,7 @@ import createRoutes from 'routes';
 import configureStore from 'store/configureStore';
 import preRenderMiddleware from 'middlewares/preRenderMiddleware';
 import header from 'components/Meta';
+import ssrAuth from 'api/preRenderAuthentication.js';
 
 const clientConfig = {
   host: process.env.HOSTNAME || 'localhost',
@@ -34,6 +35,7 @@ export default function render(req, res) {
   }, history);
   const routes = createRoutes(store);
 
+
   /*
    * From the react-router docs:
    *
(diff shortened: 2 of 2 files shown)


choonkending:
Thanks for taking the time to write up the description! I'll test it tonight!

choonkending:
Linking to the original issue https://github.com/choonkending/react-webpack-node/issues/171

choonkending:
Thanks for that!

Found a minor bug though 👾 

**Steps to reproduce**
- Log in
- Refresh
- Notice that data fetching fails with rejected errors such as `adapter` undefined

After a bit of digging around, it was because we did not return `config` in the interceptor. So we were causing side effects but the promise did not return anything!

**Proposed Solution**

```javascript
let interceptorID;
export default function ssrAuth(cookie) {
   if (arguments.length === 0) {
     axios.interceptors.request.eject(interceptorID);
   } else {
     interceptorID = axios.interceptors.request.use(function(config) {
       config.headers['cookie'] = cookie;
       return config;
     }, function(error) {
       return Promise.reject(error);
     });
   }
}
```

**Changes**
- Return config
- Fixed indentation
- Changed `.handlers = []` to `.eject()`

I think we should change the way we eject the interceptor, so that it can delete the interceptor, following the docs [here](https://github.com/mzabriskie/axios#interceptors). 

It might be safer to use the method provided, mainly because we do not want to be changing any handlers not related to the interceptor we added - I was looking at the code [here](https://github.com/mzabriskie/axios/blob/6554493bdd3225ff87a26aa548dc562978f6d596/lib/core/InterceptorManager.js#L32) hehe.

Epochi:
I have no idea why I deleted 'return config'! 

I did not use .eject(); because it sets the interceptor to null, instead of deleting it from the handlers array, and as the page gets reloaded that .handlers array keeps growing infinately.  

I believe axios.interceptors.request.handlers array was how I noticed that cookies mix up with several requests happening at the same time, because every server prerender still uses same server-wide axios.
If we start preRenderMiddleware for one user and the second user starts it before the first one finishes, the both users will get data that's intended for second user.

It's amazing that you remembered #171!  I just walked a big circle around the same problem...
I'll have to think about this problem some more, because it seems that custom axios instance still needs to be created for server to differentiate between users. 
I'm going to be away for a week from now, should I close the PR in the meantime and while I find a solution?

choonkending:
I see, did not consider that the data could bleed between sessions!

Yep agreed, sounds like a custom axios instance is needed.

I think it's fine to leave it open so others will get pointed to this PR,
and I can let anyone who asks know that you're away :)

choonkending:
Do you think intercepting with a middleware makes more sense?

sidharthchugh:
any updates on this.

emmya:
Is this still the best fix?

choonkending:
I think #577 is a closer match to what I would like implemented

Epochi:
I personally ended up reverting to the method outlined in #184, because I haven't found a way to make this work with global axios, but thats very 'dirty' in comparison to #577 (which looks sleek af) and doesn't work with current master of reactGO

PR #227, opened 2016-05-10
Fixed typo in readme

davidalee:
Fixed a typo.

From `...we assume that you serving the page...` to `...we assume that you **are** serving the page...`.


choonkending:
Thanks for contributing!

PR #238, opened 2016-05-20
Added height: 100% to #app div

IljaDaderko:
In case if users want to use fullscreen background or videos html and body will be set 100% height, #app needs to reflect this. This practice is used in various other react starters :)

### DIFF
--- a/app/server.jsx
+++ b/app/server.jsx
@@ -85,7 +85,7 @@ export default function render(req, res) {
               ${header.link.toString()}
             </head>
             <body>
-              <div id="app">${componentHTML}</div>
+              <div id="app" style="height: 100%">${componentHTML}</div>
               <script>window.__INITIAL_STATE__ = ${JSON.stringify(initialState)};</script>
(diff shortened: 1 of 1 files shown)


choonkending:
@IljaDaderko Thanks for raising this as a PR!

Just wondering, is there a reason why we wouldn't include this in `main.css`? Or would that not work for what you want?

IljaDaderko:
@choonkending good question, to be honest I have no real reply to it, however the reason I did it as inline style is mainly because every other framework I used (even angular ones) do it this way.

choonkending:
Do you mind if we set the height in main.css? Just so we can keep it within
one place and it will be easier to maintain :)

But if that approach doesn't solve the problem for you let me know, if it
works let's update the PR :)

choonkending:
@IljaDaderko Heya, https://github.com/choonkending/react-webpack-node/commit/b2fbc0dba706a4ce02262df0d0370c31ee60939e should fix it. Let me know if it doesn't for you. Closing for now. 

Thanks for the PR!

Issue #241, opened 2016-05-24
Optimistic updates


colinlmcdonald:
### Description
- Type: Enhancement 
- Related Issue: 140

---------------------------------------------------------------
## Enhancement

**Reason to enhance/problem with existing solution**
Updates should happen on success. If there is no success, error message should happen instead.

**Suggested enhancement**
Move optimistic updates to on success updates in incrementCount, decrementCount, and destroyTopic functions in topics.js. Add .catch to all three as well along with accompanying error messaging.

Also, add testing for those functions as well stubbing out axios and the appropriate method.

-----------------------------------------------------------------

## Question
I can't see a reason for the optimistic updates as opposed to on success updates, but if you have one no problem.

choonkending:
Agreed! Thanks a lot for contributing!!

colinlmcdonald:
No worries! I'll try to have a pull request for you in the next couple of days.

colinlmcdonald:
So I added some error handling for the cases listed above and added an 'error' key to the state. I could work on mapping that to actually displaying the message and adding tests for components as well. 

Let me know if that sounds good and if you had any suggestions/opinion on how the error messaging handling should operate and which component it should go in.

choonkending:
Good work @colinlmcdonald! Thanks a lot for contributing! I'll close this issue, but feel free to re-open if you still have any enhancements/fixes related to this!

PR #247, opened 2016-05-27
Async and Reducer Tests

colinlmcdonald:
- Added async tests for increment, decrement, and destrory for both on success and on fail
- Added .catch for increment, decrement, and destroy
    - Dispatched 'createTopicFailure' as a placeholder
- Added reducer tests for all reducer cases
- Added 'error' key on state to manage error messages

Issue #241


dhyegofernando:
:+1: for this

thanks @colinlmcdonald

PR #254, opened 2016-06-02
added conditional google analytics script

caranicas:
Was adding google analytics to my project and i realized the boiler plate didn't it so I added a lightweight conditional implementation. 

It is out of your way if you don't want to use it, and super easy if you want to enable it. 

- [x] Add conditional Google Analytics script 
- [x]  Update Readme

### DIFF
--- a/app/server.jsx
+++ b/app/server.jsx
@@ -16,6 +16,32 @@ const clientConfig = {
 // configure baseURL for axios requests (for serverside API calls)
 axios.defaults.baseURL = `http://${clientConfig.host}:${clientConfig.port}`;
 
+
+const analtyicsScript =
+  typeof trackingID === "undefined" ? ``
+  :
+  `<script>
+    (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
+    (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+    m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
+    })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
+    ga('create', ${trackingID}, 'auto');
+    ga('send', 'pageview');
+  </script>`;
+
+
+/*
+ * To Enable Google analytics simply replace the hashes with your tracking ID,
+ * and move the variable above the analtyicsScript. Because the ID is declared
+ * beneath where is is being used, it's declaration being hoisted to the top
+ * of the file, however it does not maintain its value during this process,
+ * so it is undefined for the type check.
+ */
+const trackingID  = "'UA-########-#'";
+
+
+
+
 /*
  * Export render function to be used in server/config/routes.js
  * We grab the state passed in from the server and the req object from Express/Koa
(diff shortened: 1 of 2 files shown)


choonkending:
You're on fire!

caranicas:
Haha thanks. I've got a lot of spare cycles so I am just trying to do my part.

caranicas:
@choonkending I extended the comment and linked to some information about hoisting, because it seemed like a good teachable moment.  I also added some simple read me information directing the reader where to find the code and where to read more about analytics. So I think this pull request is good to go.

app/server.jsx
```diff
@@ -16,6 +16,34 @@ const clientConfig = {
...
+ * To Enable Google analytics simply replace the hashes with your tracking ID
+ * and move the constant to above the analtyicsScript constant.
+ *
+ * Currently because the ID is declared beneath where is is being used, the
+ * declaration will get hoisted to the top of the file.
+ * however the assignement  does not, so it is undefined for the type check above.
+ * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/var#var_hoisting
+ */
+const trackingID  = "'UA-########-#'";
+
```
choonkending: Would it make more sense to have this tracking ID in our [appConfig](https://github.com/choonkending/react-webpack-node/blob/master/server/config/appConfig.js) so that users can change that value without having to move the constant above the line?

choonkending: Only problem with appConfig is that it's within the `/server` directory, and this trackingID is a `clientSide` config I guess.

choonkending: I'm happy to merge as-is, and we can figure out config refactoring in the future.

choonkending:
Made a couple of comments, let me know what you think. Otherwise I'll merge and we can pull out config in the future!

caranicas:
yeah figuring out where everything goes id definitely of the tricks of doing thing  ~~isomorphically~~ universally. I don't have an issue with keeping the analytics ID in the app config, however if you put it there then I'm not sure the best way to get that value into server.jsx. Would you import it or some how pass it down as a value from the server?

The only detriment would be loosing an opportunity to explain hoisting to and actually use it in a way that works towards your advantage.

Also I'm not sure if this line is supposed to be pseudo code, but this is not going to evaluate properly.
`defaultExport.ANALYTICS = "UA-##########" || 'undefined';`

I think it would be best to pull in now, and then we can refactor once we get a second value that fits this paradigm.

choonkending:
All good, and I agree. 

You're right, I had meant to to be like
```javascript
//You'd have to export ANALYTICS_ID in your environment

defaultExport.ANALYTICS = process.env.ANALYTICS_ID;

// or just hardcode it

defaultExport.ANALYTICS = 'UA------';
```

What I had in mind was:

```javascript
// server.jsx
import { ANALYTICS} from 'appConfig';

const trackingID = ANALYTICS;

const analtyicsScript =
   typeof trackingID === "undefined" ? ``
   :
   `<script>
     (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
     (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
     m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
     })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
     ga('create', ${trackingID}, 'auto');
     ga('send', 'pageview');
   </script>`;
```

choonkending:
Merging as-is.

Good work @caranicas

mhlavacka:
`ga('create', ${trackingID}, 'auto');` should be `ga('create', '${trackingID}', 'auto');`

**'${trackingID}'**

Issue #245, opened 2016-05-25
was wondering whether react webpack node can run on aws 


elitekode:
Note: This is just a template, so feel free to use/remove the unnecessary things

### Description
- Type: Bug | Enhancement | Question
- Related Issue: `#abc`

---------------------------------------------------------------
## Bug
**Expected Behavior**

**Actual Behavior**

**Steps to Reproduce**

----------------------------------------------------------------
## Enhancement

**Reason to enhance/problem with existing solution**

**Suggested enhancement**

**Pros**

**Cons**

-----------------------------------------------------------------

## Question

**How to?**

caranicas:
I have a project that I built with this boilerplate  currently up and running on Amazon AWS. Your question of how to is a bit vague, so I will just run down the steps I took. and provide a few links. 

I set up an AWS account using the free tier. I didn't feel it was necessary to go with something like elastic beanstalk, so I used a pure EC2 instance. 

Before you make our instances you need to create a security group, which specifies what ports are open on your server, and what if any or all IPs have access to them.  The [Amazon docs](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-network-security.html#creating-security-group) can walk you through it nicely. 

Then you should make sure you are in correct region (in the upper right hand corner), and and then pic Launch a new instance, and follow along with the setup wizard. I picked the Ubuntu Server 14.04 free tier, and most of the default options and I moved through the process. And then let it set it self up. 

Once this was done I set up an elastic IP and paired it with my instance. This is important to do. because it gives you a dedicated IP for your virtual instance/IP. And make sure you pair it right away, because there is a charge for having unpaired elastic IPs.

Now it is time to connect. There are instructions on the instances page. I use SSH, with the key file that i generated in the creation process. Once you are in you need to run a bunch of updates, and then start installing. 

First install node. following the [instructions](https://nodejs.org/en/download/package-manager/)
Next install mongo db following this [instructions](https://docs.mongodb.com/v3.0/tutorial/install-mongodb-on-ubuntu/), there are a lot of them so make sure you follow the instructions for the correct version. 

To get mongo running fully was a bit of a challenge, I was having some issuing with file permisions but the notes I have look like this. This should be everything, but if something doesn't work the error and answer is on stack overflow. Because thats how I got it working. 

RUN MONGO TEST:
`sudo service mongod start`
`sudo service mongod stop`
START MONGOD:
[...]

PR #255, opened 2016-06-03
first draft of aws read me

caranicas:
Wrote out a pretty long tutorial, elaborating on my comment from issue #245. 

@slavab89 do you think you could add some information about the elastic load balancer and how to use it for the redirect? Are there any costs associated with it? And what the pros and cons of that would be vs NginX and how to install and set it up?

DevOps, (if this even merits that name) is not my forte. So this set up is just the bits and pieces  I've gathered from reading doc and stack overflow, not representative of a larger knowledge base. 

- [x]  Write a first draft
- [x] Explain Ip table
- [x] Explain better port forwarding
- [ ] Explain route53 DNS
- [ ] Link to on main read me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,6 +177,9 @@ pm2 start server/index.js
 pm2 startup ubuntu
 sudo env PATH=$PATH:/usr/local/bin pm2 startup ubuntu -u sammy
 
+
+
+
 ```
 
 Read more on DO config [here](https://github.com/choonkending/react-webpack-node/blob/master/docs/GettingStartedWithDigitalOcean.md)
--- a/docs/aws.md
+++ b/docs/aws.md
@@ -61,3 +61,6 @@ I high suggest that you do no use this method to copy over your dependencies, it
 Now that your files and dependencies are on the server a simple `npm start` will get you going. But it wont keep you going. Once you close your SSH connection your site will go down.
 
 To remedy this will will use [forever](https://www.npmjs.com/package/forever). We will need to install it globally `sudo npm install forever -g` and then add another script to our package.js. Just a modified version of the start stript. `"stayAlive": "cross-env NODE_ENV=production forever start compiled/index.js",`
+
+If you don't want the port value ruining your nice url then you can use IP tables to set it to  redirect it to 80 
+`sudo iptables -t nat -I PREROUTING -p tcp --dport 80 -j REDIRECT --to-port 3000`
--- a/docs/aws.md
+++ b/docs/aws.md
@@ -1,0 +1,63 @@
+# Deploying To AWS
+
+Amazon AWS allows you access to a free tier of service for a year, so it is a great option if you are looking for a place to host your project and want to grow your Ops skills along with your dev skills.
+
+The first thing you will need to do is sign up for an [account](https://aws.amazon.com/free/)
+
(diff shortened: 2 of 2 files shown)


slavab89:
I will not be near my working PC for a few more days so i dont have access to all the stuff + i dont want to stall you hehe, but i wrote some info and data about what you asked - feel free to correct my grammar mistakes and overall description :laughing: 
Also, there are a lot of blogs and examples over the web explaining how to do exactly all those things =]

===========================

Although it is possible to achieve the wanted result using iptables, and although it is a valid option, there are better options out there that allow more flexibility and options - for example, when using iptables approach, if one would want to protect his web server, he would need to implement that logic on his app or set up some different app configuration specified for this. Moreover, using iptables requires root permissions. Something that is not always available.

Another options out there is the good-old NginX. It is basically a web server (like Apache) that has abilities of reverse proxy that would allow our Node/React application to run on its own port but allow outside user to access it through port 80. Here is a simple tutorial that i've found can help setting this up (requires root). https://eladnava.com/binding-nodejs-port-80-using-nginx/
NginX is used in many places therefore it has lots of capabilities such as blocking ip's, detecting some web attacks and ultimately can partially help protect your application. It allows the web application to run on its own while all the network logic would be configured and running on a different process that's unrelated to the application.

Lastly, Amazon ELB service. We are running on an Amazon EC2 instance. Why not use more features and services that Amazon provides that can help us? No root? No problem!
Amazon has a service called ELB - Elastic Load Balancer - that basically performs load balancing between several instances and monitors all that allowing you to see all kinds of metrics.
Sounds cool and easy? It is, but it will cost you. As of now, the price for using it in US is: $0.025 per Elastic Load Balancer-hour (or partial hour) + $0.008 per GB of data processed by an Elastic Load Balancer. More info and prices examples can be found on their site: https://aws.amazon.com/elasticloadbalancing/pricing/
Once you're in your amazon web console, in the EC2 service. Go ahead and enter Load Balancers. From here you can create a new load balancer. You will be required to choose a unique name for the load balancer followed by some options.
**Listeners and networking** - Here is possible to set up the port redirection. On the left side port 80 should be chosen with HTTP protocol and on the right side port XXXX should be chosen, where XXXX is your application port.
**Health Check** - Its possible to define a custom health check, meaning that once every defined time a request will be sent to the server to check if its still alive. The default option will be to perform a TCP port check on port XXXX. This will just check that the server is alive. But its also possible to send an actual HTTP request that the web application could receive and then return a status code that will indicate if there is a problem on the server or not. For example, if the application uses a DB but is unable to connect to it for some reason then the ELB can know about it and take that server out of its instance list.
**Instances** - Add your EC2 instance that runs the application.
Once all that is set up, you can go over to the instances tab and see that the instance is in a Running state, meaning that everything is well and you can access your application through the ELB.

For the last part, we would want to run under some domain name instead of a long ELB name (that can be seen on the description of the ELB). For that we would use another amazon service called Route53.

caranicas:
@slavab89  thanks a lot! That link about NginX was really informative, and I am definitely going to convert my project over to that. 

It really is pretty simple to implement, and it seems like a much better option than IP tables.  

ELB looks interesting but I think it might be overkill, and as a pay service i'm not sure that is the best thing to recommend from this boilerplate.

PR #256, opened 2016-06-03
Docs/egghead

caranicas:


caranicas:
@choonkending watching these videos taught me a ton, and got me up to speed so fast. I think they are definitely worth linking to in the read me.

choonkending:
Agreed! I was just going to spend my weekend watching them again lol

PR #257, opened 2016-06-04
transform Object.assign for browsers lacking support

### LINKED ISSUE #243: Uncaught TypeError: Object.assign is not a function
### Description
- Type: Bug

---------------------------------------------------------------
## Bug
**Expected Behavior**
Using Google Chrome 44, I should be able to type something in the `Vote for your top hack idea` textbox without JavaScript errors. Furthermore, I should be able to login and register.

**Actual Behavior**
Whenever I execute any of the actions stated above, I get the JavaScript error `Uncaught TypeError: Object.assign is not a function`, and I cannot register nor login.

**Steps to Reproduce**
Visit: https://react-webpack-node.herokuapp.com/ using Google Chrome 44 (Linux Mint 15) with experimental Javascript features turned off. Type something in the bar below `Vote for your top hack idea`, check the console.

**Root Cause and Suggested Solution**
I found that the error is caused due to the fact that my Google Chrome version does not support ES6. As such, I propose to add an ES6 polyfill to the project. See: http://www.zsoltnagy.eu/using-es6-modules-with-webpack/

psimyn:
fixes #243

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     "babel-loader": "^6.2.4",
     "babel-plugin-react-transform": "^2.0.2",
     "babel-plugin-transform-decorators-legacy": "^1.3.4",
+    "babel-plugin-transform-object-assign": "^6.8.0",
     "babel-plugin-transform-react-constant-elements": "^6.5.0",
     "babel-plugin-transform-react-inline-elements": "^6.6.5",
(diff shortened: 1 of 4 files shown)


psimyn:
@choonkending merging this one ok? Also probably merging the postInstall PR :)

choonkending:
:ship:

psimyn:
🙇

PR #259, opened 2016-06-04
use setup script instead of postInstall

### LINKED ISSUE #167: npm install for a second time
There seems to be an issue if you re run "npm install" without specified package name. I pulled the new version with nodemon and  got in running. the first thing I did was move over my dependencies and go to install them. However when i did this. it seems to have kicked off the whole rebuild process.

It isn't the biggest issue, because normally I add new packages one at a time, so it is a low priority but probably an easy fix. 

screen shot 2016-03-11 at 1 04 55 pm

screen shot 2016-03-11 at 1 05 13 pm

psimyn:
means you aren't running full prod build when following dev steps, saving $$$
fixes #167

### DIFF
--- a/docs/deployment/Heroku.md
+++ b/docs/deployment/Heroku.md
@@ -4,6 +4,9 @@ Heroku
 ```bash
 heroku create
 
+# Run build after npm install
+heroku config:set BUILD_ASSETS=true
+
 # Deploy to Heroku server
 git push heroku master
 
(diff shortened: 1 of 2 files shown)


choonkending:
Might have to test on heroku

choonkending:
https://devcenter.heroku.com/articles/node-best-practices#hook-things-up

:( the demo might not work. Should I then merge then and figure a way for heroku deployment?

psimyn:
using environment vars will work, pretty much exactly like their example

```
"postInstall": "if [ $BUILD_ASSETS ]; then npm run build; fi"
```

I'll update tomororw

psimyn:
@choonkending can you test heroku with this change?

choonkending:
@psimyn Sorry, just tested, doesn't work. Heroku doesn't build it :( I wonder why.

Issue #136, opened 2016-02-11
export constants from actions?


charlieclark:
More a suggestion than an issue -
It feels a bit more organized and DRY to export the actions constants from the action itself.
eg: https://github.com/caljrimmer/isomorphic-redux-app/blob/master/src/common/actions/reddit.js

choonkending:
@charlieclark I have been thinking about this myself. I think it's an excellent idea, because then the when we import it in reducers as well, it would actually give the user more context as to where the action came from.

slavab89:
This is a pretty old issue and we're currently in a process of cleaning up the project's issues. Quite a lot of things have changed since this issue was raised. If you still think this is relevant, feel free to comment here or open a new issue.

PR #260, opened 2016-06-04
Shuffle [WIP]

psimyn:
Started component-y refactor. Let me know if you hate it an I'll abort, otherwise not too much left to go. Goals of this are standalone sections (all related compoents/actions etc are self-contained). This will make a "clean" version easier (issue #156), keep constants near relevant code (#136), and be easier to add a 'generate' style task to add a new component/page.

Things left to do:
- [x] everything in components folder
- [x] few more containers
- [ ] move store/middleware to a `redux` subfolder??
- [ ] common / toplevel stuff to 'common' folder??
- [ ] refactor / fix tests

Sources/ideas/explanations (angular one is from 2 years ago!)
http://jaysoo.ca/2016/02/28/organizing-redux-application/
http://marmelab.com/blog/2015/12/17/react-directory-structure.html
https://scotch.io/tutorials/angularjs-best-practices-directory-structure


caranicas:
if you are taking the time to shuffle the project around then I would like to propose a new folder structure for the react components. I've modified my personal project work this way and I think it is a good way to think about it. 

The first thing I did was replace top level files with folders of the same name and an `index.jsx`  instead. I think this makes things easier to think about and gives you a natural place to extend the component.

The first thing is that I move everything into a `display` folder. With 4 sub folders `chrome`,  `components`, `modules`,  and 'pages'. 

- `chrome` is or the global pieces and things that make up the application frame.
- `components` are all of the single function display components. buttons, inputs, things like that.  
- `modules` are composed of multiple components. 
- `pages` are files that map to routes

Once you get to the page files there are some more sub folders, `wraps` and  `modules`. 

- `modules` is for any thing that is composed of multiple components but has page level specificity.
- `mapped` is for page specific components or modules that then need to wrapped in a  `connect` function with redux. 

At this level I have stopped the practice of named folders with `index.jsx` because with things being page specific because if there is an opportunity to generalize it should first be moved into the component or module directory. 

This is a simple example demonstrating what I am talking about. I'd love to here your opinion and @choonkending 's

```
app
- display

    - chrome
        - Nav
            - index.jsx
        - App.jsx
        - Meta.jsx

    - components
        - Button
            - index.jsx
        - Login
            - index.jsx
        - Register
            - index.jsx

    - modules
        - LoginOrRegister
            - index.jsx

    - pages
        - LoginPage
            - mapped
                - WrapedLoginOrRegister.jsx
            - modules
                - LoginButton.jsx
            - index.jsx
```

psimyn:
@caranicas thanks for the ideas and feedback! I really like the chrome/components breakdown; was unsure what to do with top-level stuff like App and Meta. And definitely in favour of the components breakdown for simple/common components - I don't think are are any this low-level in the repo yet, but I can add some placeholders so the proposed structure is clear.

Few questions:

- Should single use modules live only in their relevent Page folder? For example, `LoginOrRegister` is unlikely to be used outside of a LoginPage
- Is multiple `connect`ed components in a single 'page' common practice? Or should that be considered the exception?
- what is the need of the `display` folder? Are there other folders at that level or is the idea to split out data-fetching etc?

My main goal of this is Adding a new page/feature should be as self-contained as possible. So when people say "I want this repo without the Users bit" or "without the topics page", you should just be able to delete the folder and it's done

choonkending:
@caranicas I like the idea of having `pages`, `components`, and `modules`. Having top-level stuff in `chrome` is good but I think that:
- we should rename it to something not too specific, e.g. `root`? lol
- Move `meta` things into `pages`. Will explain below. 

I also like that you have a `mapped` separation with the actual module. But I share @psimyn 's question with 

> Is multiple connected components in a single 'page' common practice? Or should that be considered the exception?

If there is only one connected component at that level, then we could have `LoginOrRegisterContainer` and `LoginOrRegister` exported via the `index.js`

We've been experimenting with module structures at work recently, and kind of like
```
- pages
  - meta specific static methods - doing data fetching
- handlers
  - route specific configuration
- components
   - LoginOrRegister
       - index.js
       - LoginOrRegisterContainer - wrapped with `connect`
       - reducers.js
   - Login
      - index.js
```

I like keeping reducers within the components @psimyn. 

I'm still sitting on the fence on this one, and it boils down to this:
- Should we keep all our components (dumb and smart ones) within the single folder `components` and expose them via an `index.js`. This would be similar to combining @caranicas 's `modules` and `components` under a `component`.
  - Advantage: Makes sense because they are all **components**.
  - Disadvantage: Humongous God folder

- Should we separate them into `modules` and `components`?
  - So `modules` will be exposing `reducers, containers`?
   - Advantage: Clear separation of pure components that do not mess around with state
   - Disadvantage: Might be a *bit* confusing.

I've tried my best to condense my brain dump. Hopefully I understood both @psimyn @caranicas 's motivation and am replying like a logical human being. If so, feel free to point out that I missed the point.

caranicas:
A lot to respond to so  ill try to break it down 

@psimyn 

- Glad you liked the idea behind `chrome`, i struggled for a long time before i settled on that. 

- Yeah and there might not be some sufficiently low level components. but I think I have a few basic react wrapped text inputs lying around that I could use to demonstrate this.

- Yes you are definitely correct about `LoginOrRegister` being single use and it would make more sense in this case. I think i was just looking for a way to exemplify using pieces in the boilerplate. 

- In terms of multiple connected components I'm not sure what the "best practice" is, but it is how I've been building my projects. Maybe I misunderstood, but I thought that was the whole idea behind high order components and composition. 

My usual flow involves building a small pure component, and then generalized modules, and then page specific modules. And this is where I think it gets interesting. The pattern I've fallen into with these is to first wrap it in a `mapDispatchToProps`, and then if there are more then one of these modules on the page,  they all get their own wrapped version with a 'mapStateToProps`. This allows my pages to stay very slim. Just the top level page structure defining the layout and the modules therein.  Using this I have found it really easy to iterate on new layouts and experiment. 

- the point of the display folder is just to group the sub folders. I don't know how I like 'chrome', `components`, `modules`, and `pages` being on the same level as `reducers` and `actions`. If you want to name is `root` that is fine with me, just as long as it contains all the display  "logic"

- As for reducers I don't think they should really be grouped with the modules. I don't understand the motivation behind that move. It has not been my experience that reducers map to containers, but manage app state, so I think that might be misleading.

**HOWEVER** 

If you are talking about creating page level reducers to manage page level state like what is focused or being hovered over then that is some next level craziness that I could get behind. Currently I had been managing things like that within my `mapPropsToState` function, which works nicely but can get a bit unwieldy. 

On a side note, this is awesome.

choonkending:
@psimyn Could you rebase and merge? :)

app/users/reducer.js
```diff
@@ -0,0 +1,92 @@
...
+  TOGGLE_LOGIN_MODE,
+  MANUAL_LOGIN_USER,
+  LOGIN_SUCCESS_USER,
+  LOGIN_ERROR_USER,
+  SIGNUP_USER,
+  SIGNUP_SUCCESS_USER,
+  SIGNUP_ERROR_USER,
+  LOGOUT_USER,
+  LOGOUT_SUCCESS_USER,
+  LOGOUT_ERROR_USER } from './actions';
```
choonkending: Are moving towards explicitly importing the required actions?

I like the idea of exporting constants within the `actions.js` file, but that can quickly become a large file. Perhaps we could still keep `types` within each module.

```javascript
component/
     -reducers
     - types
     - actions
     - index
     - components/
         - Container
```

psimyn: this was mainly to speed up my resolving merge conflict, happy to change it back.

Main reasons for keeping them within actions:
 - they are constants used to describe actions
 - generally all needed by actions
 - saving the 'type' label for adding type annotations :)

choonkending:
> If you are talking about creating page level reducers to manage page level state like what is focused or being hovered over then that is some next level craziness that I could get behind
I agree with this.

I have seen good use cases of reducers within a `page` level component that makes sense to sit within that folder, so that you could easily reuse the entire component in a different project with that reducer exported. The main motivation for this in projects I've been involved in is reusability, particularly if that component is used in different projects altogether.

However, we need to be aware with having reducers in components, and make sure that we don't just create a reducer for every component there is. 


Redux or flux in particular is awesome because the state in your store isn't necessarily mapped to your component, and you can easily derive the data to be used within your component without changing the store. 

So far @psimyn changes look safe, as `messages`, `users` are all page level state, and seems okay for me.

Ideally we should keep our store pure, and each component would have a `selector` that derives computed data from our redux stores.

choonkending:
Closing this as it's been a year, and I think we all might have different opinions about what we had a year ago! 😬  Happy to discuss this in an issue!

PR #269, opened 2016-06-10
[SPLIT #221]  Updated frontend code for new components

18601673727:


caranicas:
@choonkending any movement on this one #270 or #272 ? I've looked at the diffs and the look pretty good, definitely some steps in the right direction. 

@psimyn have you had look ?

choonkending:
@caranicas @18601673727 Apologies for the late review. I looked at #270 and merged it (but later found it caused a couple of bugs) so I pushed a revert.

Planning to do a batch review the end of next week and get all the goodies merged in 👍

PR #271, opened 2016-06-10
[SPLIT #221] Redux axios and loading bar

18601673727:


svrcekmichal:
Hi @choonkending , as an author of original [redux-axios-middleware](https://github.com/svrcekmichal/redux-axios-middleware) I just want to warn you, that @18601673727 is attempting to merge into your project his fork of middleware called [redux-axios](https://github.com/18601673727/redux-axios). I don't want to sound bad, but I just don't want your work to get destroyed and to repeat this process of middleware selection few weeks later when you found out that forked middleware isn't developed and maintained anymore.

choonkending:
@svrcekmichal Thanks for letting me know buddy! I was not aware of this until you pointed it out to me. No you don't sound bad at all. I'll keep this in mind when I get the chance to review the PR.

18601673727:
Narrow minded people can always predict the future, can't they? 

Just wanna to clear the air a little bit, `redux-axios` now is a fork of `redux-axios-middleware`, but `redux-axios` is designed to be a redux abstraction of axios(http client), not like `redux-axios-middleware` the bridge way, you don't have to new serval `axios` instances in `configureStore.js` or have a `"axios": "x.x.x"` in your `package.json`, to achieve the same thing.

`redux-axios` want to reduce users' code as much as possible. But it looks like someone pissed off, so maybe I'll rewrite it someday, not just a fork plus modification. 😄

choonkending:
Whoa that is a bit of a heated discussion going on here!

Here are my comments purely based on the code and changes in this PR.

Here's what I like:
- Moving `configureStore.js` out and removing the `store/` directory
- Creating a `client` index and backend module that makes it easy to change configurations
- `res.sendStatus(500)` change in `server.js`

I disagree with the addition of `react-redux-loading-bar` and `redux-axios` (still) unfortunately. We've had the redux-axios middleware discussion prior to this, and while it definitely is beneficial for many projects, I think that _cost_ of adding this abstraction to this specific boilerplate outweighs the benefit. 

I'm looking for ways to reduce the power of our code in a principled way https://github.com/choonkending/react-webpack-node/issues/364

I'm going to close this PR off, but will be happy to get the above changes as PR(s) or I can apply them when I am _reducing_ this boilerplate. Sorry we did not get to merge this, and even though we did not apply this change, I think it still spurred the discussion regarding _abstraction_ and _side effects_ in the right direction as this boilerplate matures.

PR #342, opened 2016-07-16
Update nodemon.json

ZeroCho:
I excluded **webpack.config.prod.js** from **nodemon.json** because it doesn't need to be watched in dev mod.

Sometimes it's annoying that the server restarted when I updated webpack.config.prod.js


choonkending:
@ZeroCho Thanks! This makes sense. Merging!

Issue #261, opened 2016-06-04
Unit Testing Stateless Components w/ Karma


jrodl3r:
Having trouble unit testing even the simplest of stateless components. Getting it to fail/not-render is no problem, just passing the rendering test w/ good array data.

Any gotchas that I might be missing on wiring-up tests for stateless component?

Thx

choonkending:
Need to use react-stateless-wrapper for testing I believe

jrodl3r:
**`/app/tests/components/MainSection-test.js`**

```
import { wrap } from 'react-stateless-wrapper';
import MainSection from 'components/MainSection';
import TopicItem from 'components/TopicItem';
...
const WrappedMainSection = wrap(MainSection);
...
describe('Has topics', () => {
    it('should render TopicItems', () => {
      result = ReactTestUtils.renderIntoDocument(<WrappedMainSection topics={topics} {...stubFunctions} />);
      topicItems = ReactTestUtils.scryRenderedComponentsWithType(result, TopicItem);
      expect(topicItems.length).toBe(1);
    });
  });
```

How would _this_ test work/change if the `TopicItem` component were stateless?

Thx

choonkending:
You would need to wrap TopicItem in a `wrap`.

```javascript
const WrappedTopicItem = wrap(TopicItem);
```

Anything stateless in react needs to be wrapped in a component. 

We might consider moving towards [Enzyme](https://github.com/airbnb/enzyme) from airbnb in the future if we find this not a good idea.

jrodl3r:
I get that it needs to be wrapped, but on this MainSection-test example, the `TopicItem` is being passed to `ReactTestUtils.scryRenderedComponentsWithType(result, TopicItem);` - If you change that `TopicItem` param to `WrappedTopicItem` the test fails.

I guess I'm asking how you can have a Karma test here where both the parent and child component being tested are stateless. Does that make more sense? Thx!

jrodl3r:
Would love to setup a PR to add Enzyme support for UI testing... However, it's a little fuzzy for me as to how you would add Enzyme to the current testing configuration (along side the _actions_ and _reducers_ tests), while removing the current "wrap" system that is in place for testing UI and rendering.

Point me in the right direction and I'll be more than happy to take a shot, Thx!

jrodl3r:
[...]

PR #344, opened 2016-07-16
Enzyme (React Component Unit-Testing)

jrodl3r:
Re: #261

### DIFF
--- a/app/tests/components/MainSection-test.js
+++ b/app/tests/components/MainSection-test.js
@@ -1,43 +1,33 @@
 import React from 'react';
-import ReactTestUtils from 'react-addons-test-utils';
 import expect from 'expect';
-import { wrap } from 'react-stateless-wrapper';
+import { mount } from 'enzyme';
 import MainSection from 'components/MainSection';
-import TopicItem from 'components/TopicItem';
 
-const WrappedMainSection = wrap(MainSection);
+const emptyData = [];
+const topicItemData = [{
+  text: '',
+  id: '',
+  index: 0,
+  onIncrement: () => {},
+  onDecrement: () => {},
+  onDestroy: () => {}
+}];
+const stubFunctions = {
+  onIncrement: () => {},
+  onDecrement: () => {},
+  onDestroy: () => {}
+};
 
-describe('MainSection', () => {
-  let result;
-  let topicItems;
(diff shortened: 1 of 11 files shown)


app/tests/components/MainSection-test.js
```diff
@@ -1,43 +1,33 @@
...
-
-  describe('Has topics', () => {
-    it('should render TopicItems', () => {
-      result = ReactTestUtils.renderIntoDocument(<WrappedMainSection topics={topics} {...stubFunctions} />);
-      topicItems = ReactTestUtils.scryRenderedComponentsWithType(result, TopicItem);
-      expect(topicItems.length).toBe(1);
+describe('<MainSection />', () => {
+  describe('With Topics', () => {
+    it('should render <TopicItem> list items', () => {
+      expect(mount(<MainSection topics={topicItemData} {...stubFunctions} />).find('.list li').length).toBe(1);
```
choonkending: Minor comment: should this be `find(TopicItem)` given that we are asserting that `MainSection` contains `TopicItem`? Finding `.list li` works, but that ties this test to the specific implementation of `TopicItem`, i.e. if at some point in the future, we change TopicItem to be `<div>`, this test will break even though it shouldn't. :)

choonkending:
@jrodl3r Excellent changes - and thanks for pushing this through!

app/tests/components/MainSection-test.js
jrodl3r: yes, Enzyme has very flexible/expressive selectors. Do you have any suggestions for the choice to go with here? Also, is there any documentation that you would like me to add/update? Thx 👍

choonkending: I'd still go with `find` , the way you did it, following [this](https://github.com/airbnb/enzyme/blob/master/docs/api/shallow.md) doc here in enzyme.

so instead of using `.list li`

I would 
```javascript
expect(mount(<MainSection topics={topicItemData} {...stubFunctions} />).find(TopicItem).length).toBe(1);
```

>  is there any documentation that you would like me to add/update?

So far we don't have any docs for testing yet, but I think we should outline good testing practices here as a doc after this gets merged in.

Thanks!

jrodl3r:
oops... I merged before I pushed =(

PR #399, opened 2016-08-01
Small changes to the README

christiearcus:
A couple of teeny changes to the README so the community can skim through and quickly get an idea  🚤


choonkending:
Thanks @christiearcus!

PR #405, opened 2016-08-01
Fix undeclared variable in dev mode

choonkending:
# Reason
https://github.com/choonkending/react-webpack-node/pull/400 introduced a bug for dev mode


choonkending:
Self 🚢

Issue #372, opened 2016-07-25
"Remove" unneeded code for stateless boilerplate


choonkending:
### Description
- Type: Enhancement
- Related Issue:  #137 #156
## Enhancement

Was looking an [create-react-app](https://github.com/facebookincubator/create-react-app#converting-to-a-custom-setup) and was inspired by the _**No Lock-In**_ idea.

**Problem**
There is still a lot to strip out when cloning this boilerplate. 

The code that sits in this repository are **_useful_** examples of how one would implement certain features with the best practices - we strive for that.

However, I still find myself stripping out a lot of code I don't need.

We've already done a decent job with **abstracting out the DB connection** in https://github.com/choonkending/react-webpack-node/pull/176 and https://github.com/choonkending/react-webpack-node/pull/190

**Suggested enhancement**
Have an `npm run stateless` that:
- Runs`rimraf` to remove all the unneeded:
  - DB related and controller files - leaving only the server-rendering bit left
  - Front end code that gets in the way of development (this will involve refactoring a bit of our component code perhaps)

If this gets shipped, and it works swimmingly well, I reckon we can bump our version to 2.0! Yay!

**Pros**
- Easy to start work immmeeeeddiatellyyy
- Feel like a thug

Including
@caranicas @psimyn @jrodl3r @dzotokan and whoever has thoughts!

psimyn:
alternative is to strip it down to be stateless, then move example stuff to a separate repo or a branch

jrodl3r:
"The simplicity is the ultimate sophistication."

simplify _master_, setup feature branches for add-ons (Mongo, Vote-App, Todo-App, etc.)

personally, the first thing I did was flatten `/db` and remove all non-Mongo stuff... :neckbeard:

Or... Setup an ORG and do repo's instead of branches 👍

caranicas:
Ok, apparently I closed my window before hitting the comment button earlier. :expressionless: 

So enjoy my now less thoughtful comments. I agree with @jrodl3r and @psimyn. I would going to suggest the multiple repos, and allowing the stateless version to live upstream of the more tutorial oriented version.

I am embarrassed to admit how long I had the todo code sitting inside my app for reference, so I think it is indispensable resource. There are myriad boilerplates out there that focus on getting you up and running, but most of them are blog fodder. They don't focus on best practices, helping you understand how you should be coding, and the certainly aren't getting better. I evaluated quite a few before I settled on this on. 

[...]

PR #406, opened 2016-08-01
Release stateless 2.0 [ WIP DO NOT MERGE ]

choonkending:
Related issues
https://github.com/choonkending/react-webpack-node/issues/372


app/containers/App.jsx
```diff
@@ -16,18 +14,15 @@ const cx = classNames.bind(styles);
...
-  return (
-    <div className={cx('app')}>
-      <Navigation />
-      <Message />
-        {children}
-    </div>
-  );
-};
+const App = ({children}) => (
+  <div className={cx('app')}>
```
choonkending: @psimyn @jrodl3r @caranicas do you guys like this bit of coupling to `classnames?`

Perhaps this should suffice?
```javascript
const { app }  = styles;

...
className={app}
```

I'm definitely a fan of `classNames` and see benefit in using it, but as a minimal boilerplate it might be a bit too much and gives that bit of overhead in understanding how this app works.

psimyn: it's one of the most used react libs, I think better to keep it explicit. Can remove `/bind` and cx though, and just use classnames directly

when there are no conditions on a class though you can just include it directly, same as your example or import whole css module

```
import styles form './styles'

...
className={styles.app}
```

caranicas:
Yeah I agree that `classnames` is pretty ubiquitous and should be included. Attaching multiple classes on a element is a basic behavior so I think a way to concat classes is a necessity. 

`bind` becomes something that you need to start worrying about once your app state starts to influence presentation layer. 

from the `classnames` docs 

> If you are using css-modules, or a similar approach to abstract class "names" and the real className values that are actually output to the DOM, you may want to use the bind variant.

```
let className = cx({
      base: true,
      inProgress: this.props.store.submissionInProgress,
      error: this.props.store.errorOccurred,
      disabled: this.props.form.valid,
    });
```

So this would be a make sense that this improvement / increase in complexity would live down stream.

choonkending:
Closing in favor of v2 coming out!

PR #415, opened 2016-08-04
Offline support [WIP]

psimyn:
This adds a service worker that caches `app.js`. Something weird is happening with self-signed cert I tried, so insecure for now.
## Setup

Easy mode:
Launch chrome/chromium with insecure flag: `chromium --unsafely-treat-insecure-origin-as-secure`

Hard mode:
1. Generate cert and key
`$ openssl req -nodes -x509 -newkey rsa:2048 -keyout key.pem -out cert.pem -days 365`
2.  Uncomment https stuff in `server/index.js`
3. Go to https://localhost:3443/
4. ???
## Todos
- [ ] Some kind of html page cached
- [ ] Cache API responses
- [ ] Cache images
- [ ] Fix ssl dev issues
- [ ] App manifest
- [ ] Look into https://github.com/gmetais/sw-delta

### DIFF
--- a/app/server.jsx
+++ b/app/server.jsx
@@ -113,6 +113,11 @@ export default function render(req, res) {
               ${header.link.toString()}
             </head>
             <body>
+              <script>
+                if ('serviceWorker' in navigator) {
+                  navigator.serviceWorker.register('/sw.js')
+                }
+              </script>
               <div id="app">${componentHTML}</div>
               <script>window.__INITIAL_STATE__ = ${JSON.stringify(initialState)};</script>
               ${analtyicsScript}
--- a/package.json
+++ b/package.json
@@ -46,7 +46,8 @@
     "redux-mock-store": "1.1.2",
     "sinon": "^1.17.4",
     "stylelint": "^6.9.0",
-    "stylelint-webpack-plugin": "^0.2.0"
+    "stylelint-webpack-plugin": "^0.2.0",
+    "sw-precache-webpack-plugin": "^0.3.1"
   },
   "engines": {
     "node": "5.x"
--- a/server/index.js
+++ b/server/index.js
@@ -8,6 +8,21 @@ import routesConfig from './config/routes';
 const App = require('../public/assets/server');
 const app = express();
 
+// add service worker
+import path from 'path';
+import fs from 'fs';
+import https from 'https';
+
+// https.createServer({
+//   key: fs.readFileSync('./key.pem'),
+//   cert: fs.readFileSync('./cert.pem')
+// }, app).listen(3443);
+
+app.get('/sw.js', (req, res) => {
+  res.setHeader('content-type', 'application/javascript');
+  res.sendFile(path.join(__dirname + '/../public/assets/sw.js'))
+})
+
 /*
  * Database-specific setup
  * - connect to MongoDB using mongoose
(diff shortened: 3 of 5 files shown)


caranicas:
@psimyn i was confused about this, but then I happened to watch this video [Instant Loading: Building offline-first Progressive Web Apps](https://www.youtube.com/watch?v=cmGr0RszHc8) from google io.

@choonkending  @jrodl3r  if you haven't seen it is really interesting.

app/server.jsx
```diff
@@ -113,6 +113,11 @@ export default function render(req, res) {
               ${header.link.toString()}
             </head>
             <body>
+              <script>
+                if ('serviceWorker' in navigator) {
+                  navigator.serviceWorker.register('/sw.js')
```
choonkending: `;`? - for consistency mainly. I don't actually mind the absence of `;`

choonkending: Should this script be placed at the beginning of the body tag?
I've been around hunting for examples, but the service worker script mostly gets bundled with the site scripts (take a look at [Google IO](https://events.google.com/io2016/)) I believe at the bottom of the page?

psimyn: Can definitely be moved - possibly shouldn't be inlined at all and be part of app instead - this was just a dumping ground for now. Same as the catchall runtimeCaching.

Will make some more updates soon, I'll aim to get it into a mergeable state

psimyn:
@caranicas yeah I didn't explain in issue particularly well. Just wanted to create issue so I'd actually work on it :)

choonkending:
@psimyn Nice work! Definitely a good step forward for delivering a better offline experience! 👍 

[sw-precache](https://github.com/GoogleChrome/sw-precache) looks like an awesome service worker for generating service worker precaches. [sw-precache-webpack-plugin](https://github.com/goldhand/sw-precache-webpack-plugin) looks awesome as well.

```
 ASSET                        BUILD                                  NODE
[/sw.js] < ------ > [sw precache webpack plugin] < -------> [sw precache]
```
However, I'm a bit concerned that we have two layers of abstraction in order to achieve what we want - service workers for precaching. If any one of them breaks (which is unlikely but still a possibility) or get outdated, it gets harder for us to maintain.

I might invest some time this week to learn more about offline experiences (how have i been so blind my entire life?!), and see whether it's possible to come up with a less feature complete (possibly less powerful) but simpler beginner file.

**For further reading**
There is alternatively an [offline plugin](https://github.com/NekR/offline-plugin) for webpack

caranicas:
@choonkending yeah i feel you. This whole a11y phenomenon (ugh..  accessibility, did I really just abbreviate that?) is knew to me. It my day it was graceful degradation, now, its progressive enhancement. But I get it. Smarter to build something solid from the ground up rather than go back and shoe horn alternate functionality for edge cases. 

watching the video,  it is very possible to keep the initial service worker very simple. During the demo he iterates on the functionality several times. The first "shippable" version is just a custom no connection screen "owning the app state"

psimyn:
Main reason for using sw-precache was just a zero-config option for getting static assets cached. The [service worker it includes](https://github.com/GoogleChrome/sw-precache/blob/master/service-worker.tmpl) is pretty basic, and we can easily enough move that into config and treat the service worker more like the rest of the app code. Could do that anyway with the includeScripts, so it's just whether we want the additional boilerplate or not.

But having just seen [sw-delta](https://github.com/gmetais/sw-delta) I am probably pivoting the PR more toward that...

Feel free to push to this branch if you have any ideas

Issue #418, opened 2016-08-04
Secure / Encrypt Fetched-Data


jrodl3r:
Seeking solution for securing/encrypting fetched-data.

even after enabling HTTPS, all fetched-data is still raw-accessible in the console as _{'human-readable': 'response-data'}_.

Ideally, fetched-JSON response would arrive via https protocol, encrypted to keep from peering eyes.

choonkending:
Not an expert in this area, but definitely a fan of improving our security.

Do you want a

```
Encrypt in browser -> Decrypt in server

Encrypt in server -> Decrypt in browser
```

sort of model? I have 0% knowledge in this, but we could work to find out what a good way of doing this is.

**On a side node**
We should add helmet into our express to improve security.

**References**
[Security best practices](https://expressjs.com/en/advanced/best-practice-security.html)

ZeroCho:
I'm running a blog based on reactGo, and I think we need to add data-encryption and prioritize it because initial data is shown to everyone who knows how to use browser console.

For example, my blog has a problem that members' emails are scraped by search engine and robots. This is not what my members want.

One more thing. I added helmet into my blog and it works well. Helmet protects from various threats and it's very simple to use.
`app.use(helmet());`

choonkending:
Agree with both your points. We should priority security as well as performance.

- [ ] https://github.com/reactGo/reactGo/issues/119 Do you wanna create helmet as a PR?
- [ ] Data encryption

I'm not sure what data encryption approaches we should take, do you have any ideas? I'll need to start researching on ways to handle this.

ZeroCho:
@choonkending #429 here is PR

jrodl3r:
Very cool adding helmet 👍 

@choonkending We should definitely add the _option_ to `[ Encrypt-on-Server » Decrypt-on-Client ]` (emphasize _optional_ because most wont need encryption). However, I'm not sure if the async-POST-data is being encrypted as it's sent from the client to the server... I will have to inspect this more and confirm - unless @ZeroCho beats me to it ; ) 🍺

choonkending:
Good work guys! I might have to handball this to @jrodl3r and @ZeroCho (and whoever is interested) as I'm focused on improving offline support and performance for this and next week!

jrodl3r:
[...]

PR #502, opened 2016-09-11
data encryption

ZeroCho:
#418

And I think we need to make **app/config** folder for configuration on client side
@jrodl3r @choonkending @caranicas @psimyn


choonkending:
Awesome!

Yep I've also been thinking of a client side config.

I have 2 branches I want merged first :
- meta tag refactor (with react - helmet)
- fetch data on route change 

Both of these branches make significant folder changes so I don't want to step on anyone's changes. 

I wonder whether this as a performance impact.

ZeroCho:
@choonkending Okay. I'll wait until you merge those two branches and change my codes after it

jrodl3r:
I'll start testing this on my end

jrodl3r:
@ZeroCho This works like a charm so far on latest build - Nice job man! that technique you used to encrypt/decrypt was clutch as well 👍

jrodl3r:
just noticed error on `test` after adding encryption to GET + reducers:

```
ERROR in ./app/images/favicon.png
Module parse failed: /Users/xxx/app/images/favicon.png Unexpected character '' (1:0)
You may need an appropriate loader to handle this file type.
SyntaxError: Unexpected character '' (1:0)
```

jrodl3r:
@ZeroCho what about just adding a Heroku config-var and doing this (instead of modifying `helmconfig.js`)?

```javascript
const encryptKey = process.env.ENCRYPT_KEY || 'XYZ789';
```

I suppose you could stuff that into the helmconfig, or another config file - did you mention that?

ZeroCho:
@jrodl3r I also think that putting encrypt_key in helmconfig.js is not appropriate. So I think it will be good if there is a separate config file.
And about the error, I'll test it. Anyway, I have to change this PR because this has conflicts.

jrodl3r:
@ZeroCho should we also have an enabled/disabled FLAG in the config to allow for easy on/off _[off by default]_ control of encryption?

ZeroCho:
@jrodl3r made appConfig files and modularized crypto-js to decrease size. Could you please review it?

jrodl3r:
Looks good, will verify ASAP ; )

jrodl3r:
also wondering if we can use:
`webpack.DefinePlugin({ __CRYPT_KEY__: process.env.CRYPT_KEY })`

to store the key as a Heroku config-var... and then inside appConfig:
`export const encryptKey = __CRYPT_KEY__ || 'ReactGo';`

jrodl3r:
getting `SyntaxError: Unexpected token export` on `appConfig.js` for PROD Server

ZeroCho:
@jrodl3r `SyntaxError: Unexpected token export` might not be this PR's problem. It's webpack's problem I guess.

`webpack.DefinePlugin({ __CRYPT_KEY__: process.env.CRYPT_KEY })`
I think it's a good idea.

jrodl3r:
Ok. the problem I was having in `server.jsx` was because I was missing the quote tick marks on the send(`...`), so using this:
```javascript
const stateData = encryption
  ? "'" + AES.encrypt(JSON.stringify(initialState), encryptKey).toString() + "'"
  : JSON.stringify(initialState);
```
and then removing the single quotes here:
```javascript
...<script>window.__INITIAL_STATE__ = ${stateData};</script>...
```

ZeroCho:
@choonkending I think this should be prioritized due to XSS possibility
[Link](https://medium.com/node-security/the-most-common-xss-vulnerability-in-react-js-applications-2bdffbcc1fa0#.k39qmqizd)

choonkending:
@ZeroCho Most definitely! Good point ! Added to the project wall!

PR #505, opened 2016-09-11
add green keeper script

caranicas:


caranicas:
@choonkending this is a one liner to run after you switch to a greenkeeper branch that will keep you cleaned up and running in one go.

package.json
```diff
@@ -18,7 +18,8 @@
     "build": "npm run clean && npm run build:prod && npm run build:node",
     "test": "cross-env NODE_ENV=test karma start",
     "test:watch": "cross-env NODE_ENV=test npm test -- --watch --no-single-run",
-    "postinstall": "npm run build"
+    "postinstall": "npm run build",
+    "keep-green": "git merge master && rm -rf node_modules && npm install && npm start"
```
choonkending: Nice work!

Why would we need to `rm -rf node_modules`?

I guess there are certain instances where the version bumps won't work without removing the node modules directory and doing a fresh install.

This is mainly to make our work easier as maintainers right?

Nice work.

caranicas:
Yeah its just for maintainers to save typing when evaluating green keeps commits. I thrash the mode modules just for safety's sake. Better safe than sorry, waiting for the installs isn't a big deal.

PR #509, opened 2016-09-13
Fetch on route

choonkending:
**Part of ** https://github.com/reactGo/reactGo/issues/487

**High Level Changes*
- Add fetching data logic to `<Route>` components
- Remove `dispatch` logic from `preRenderMiddleware`
- Remove usage of `promise` in action dispatched for asynchronous API call in redux action calls

⚠️ ~~Tests aren't passing yet - still needs to be fixed before merging.~~
⚠️ This refactor is pretty big.


app/reducers/topic.js
```diff
@@ -47,8 +32,9 @@ const topics = (
   action
 ) => {
   switch (action.type) {
-    case types.GET_TOPICS_SUCCESS:
-      return action.res.data;
+    case types.REQUEST_SUCCESS:
+      if (action.data) return action.data;
```
choonkending: This could arguably be placed in `client` and `server`. Rather than dispatch and check, we could check before dispatching.

app/server.jsx
```diff
@@ -91,19 +92,18 @@ export default function render(req, res) {
...
       // promises to resolve before returning to browser
-      preRenderMiddleware(
-        store.dispatch,
-        props.components,
-        props.params
-      )
-      .then(() => {
-        const initialState = store.getState();
+      store.dispatch({ type: types.CREATE_REQUEST });
+      preRenderMiddleware(props)
```
choonkending: This bit is duplicated in both client and server, but I think it's better than making `preRenderMiddleware` do everything.

choonkending:
Please review

@caranicas @jrodl3r @psimyn @ZeroCho + anyone who is interested!

It's a good first step I reckon. It will enable us to start refactoring in smaller chunks hopefully.

choonkending:
Countdown to self 👊 and merging - 4.5 hours.

jrodl3r:
@choonkending Can we still call `store.dispatch(...)` inside `<Route onEnter={...}>` handlers?

choonkending:
We should still be able to, given that `routes.js` is still passed a `store` instance.

choonkending:
@jrodl3r Sure! let me know if you find any side effects. I think we will be able to come up with a better solution moving forward now that we made prerendermiddleware slightly simpler.

Will merge in 1 hour :P

choonkending:
@StefanWerW good catch! Nope it's not being used anywhere!

choonkending:
Expect a fixing PR in a bit hehe

jrodl3r:
@choonkending can you explain why we want to fire `"CREATE_REQUEST"` + `"REQUEST_SUCCESS"` on every location change?

noticed that is happening on every URL (even w/ `{ data: undefined }`) and just not sure what to make of this... Please explain this a for me? Thx!

PR #523, opened 2016-09-19
Change default Heroku MongoDB (Compose MongoDB --> MongoLab)

StefanWerW:
When a new user wants to test the reactGo, by pressing the *Deploy to Heroku* Button, the default db is Compose MongoDB, and its default (and minimum) price is $18.00/Month. I thinks that's not very helpful for those new devs that just want to give a try. If we change the default db to MongoLab, the default deployment is free.

-----------------------------------------

Compose MongoDB:
herokupaid

---------------------------------------------

MongoLab:
herokufree


choonkending:
Thanks @StefanWerW !! Thanks for thinking about that!

PR #524, opened 2016-09-19
Update references

StefanWerW:
choonkending/react-webpack-node --> reactGo/reactGo

Should also update the name in app.json and package.json ?

### DIFF
--- a/app.json
+++ b/app.json
@@ -1,7 +1,7 @@
 {
   "name": "React Node Express Mongo",
   "description": "A barebones Node.js app using Express 4 with MongoDB, and ReactJS in Flux, using Webpack as a module bundler",
-  "repository": "https://github.com/choonkending/react-webpack-node",
+  "repository": "https://github.com/reactGo/reactGo",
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "react-webpack-node",
```
StefanWerW: Maybe change to ReactGo?

app.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "React Node Express Mongo",
```
StefanWerW: Change to ReactGo?

package.json
ZeroCho: Good catch!
@choonkending

app.json
choonkending: Sounds right 💃

StefanWerW:
@choonkending The name in `package.json` and `app.json`, should I update it, or you'll do it? 😄

choonkending:
It's a race! I probably wont get to it in time cause I'm prepping some slides :P

Issue #435, opened 2016-08-16
unable to get user session for get request


sidharthchugh:
Hello Guys,

I am unable to get user session when user is logged in. For example if i want to get topics for specific user. so i need value of req.user.id on topic request. But is undefined. It looks like Bug in the code.

actions/topic.js

export function fetchTopics() {
  return {
    type: types.GET_TOPICS,
    promise: makeTopicRequest('get')
  };
}

app/Containers/Votes.jsx

class Vote extends Component {
....
  static need = [
    fetchTopics
  ]
...
}

server/config/routes.js

app.get('/topic', topicController.all);

server/controllers/topics.js

export function all(req, res) {
  console.log(req.user); // undefined
  Topic.find({}).exec((err, topics) => {
    if (err) {
      console.log('Error in first query');
      return res.status(500).send('Something went wrong getting the data');
    }

```
return res.json(topics);
```

  });
}

Any Solution?

sidharthchugh:
Hello Guys,

I am unable to get user session when user is logged in. For example if i want to get topics for specific user. so i need value of req.user.id on topic request. But is undefined. It looks like Bug in the code.

actions/topic.js

export function fetchTopics() {
  return {
    type: types.GET_TOPICS,
    promise: makeTopicRequest('get')
  };
}


app/Containers/Votes.jsx

class Vote extends Component {
....
  static need = [
    fetchTopics
  ]
...
}

server/config/routes.js

app.get('/topic', topicController.all);

server/controllers/topics.js

export function all(req, res) {
  console.log(req.user); // undefined
  Topic.find({}).exec((err, topics) => {
    if (err) {
      console.log('Error in first query');
      return res.status(500).send('Something went wrong getting the data');
    }

    return res.json(topics);
  });
}



Any Solution?

ZeroCho:
Hmm thats strange. Can you console.log req.user right after login?

Also please check whether your url(both protocol and host) and axios request url are the same.

sidharthchugh:
In login and signup it works probably because they are post request. It is not working with fetch topics. 

I just clone the master branch and req.user is for topics is undefined.

ZeroCho:
ah, it is because there's user info passed to client on initial render. I think you should put
app/server.jsx
```
const store = configureStore({
    user: {
      authenticated,
      isWaiting: false,
      **userObj: req.user,**
      message: '',
      isLogin: true
    }
  }, history);
```
[...]

PR #577, opened 2016-10-03
Allow SSR to use authentication through axios instances

benhjames:
This is to solve #435 and #225, taking inspiration from #184.

The problem is that on server-side rendering, the server sends axios requests without the user's cookie set as a header at present, so the user will never be authenticated and, if the API requires you to be logged in, will fail.

To fix this, two different instances of axios are created, one on the server which sets the user's cookie header in its config, and the client one which is just the default axios instance. These are passed through to the fetchData handlers, so you can call api.get/api.post etc from your fetch handler, and it will (automagically) attach the user's cookie to requests.

This means you can finally do SSR even if your API requires you to log in. You can also now access req.user on the API too to do cool things like find objects belonging to a specific user.

(P.S. My first ever pull request - exciting!) :)

### DIFF
--- a/app/client.jsx
+++ b/app/client.jsx
@@ -7,6 +7,7 @@ import createRoutes from 'routes';
 import * as types from 'types';
 import configureStore from 'store/configureStore';
 import preRenderMiddleware from 'middlewares/preRenderMiddleware';
+import axios from 'axios';
 
 // Grab the state from a global injected into
 // server-generated HTML
@@ -15,6 +16,8 @@ const initialState = window.__INITIAL_STATE__;
 const store = configureStore(initialState, browserHistory);
 const history = syncHistoryWithStore(browserHistory, store);
 const routes = createRoutes(store);
+
+const api = axios.create();
 
 /**
  * Callback function handling frontend route changes.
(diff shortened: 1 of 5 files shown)


choonkending:
@benhjames Thanks for this! Sorry it took _soooo_ long to reply! I've been really behind all this PRs! :( I will review this when I get the chance!!

benhjames:
No worries - I can see there are a lot of greenkeepers around!

Bramas:
I was facing the same issue, was looking for the good fix and I believe this PR is just right.

emmya:
I think I love you @benhjames

slavab89:
@GGAlanSmithee What do you think about this? (Sorry for my absence... vacations and holidays :) )

The code needs to be updated to match the latest one but the general idea is to create an `axios` instance on the server side with the `cookies` from the request so that authentication would work (A session to be present).
We can set the default `axios` parameters or something since the instance itself is currently created deep down in the code

GGAlanSmithee:
@slavab89 thanks for the ping. This looks very reasonable to me. Haven´t tested it yet though, so cannot speak on it's correctness. Regarding data-fetching in general, I think in the long run we should consider using a more standard, such as using a redux middleware etc, but that is for another discussion :)

slavab89:
@benhjames would you like to rebase and resolve the conflicts and align your code with the latest codebase that we have so that we'll be able to merge?

benhjames:
Hey @slavab89 - it seems like a reasonable amount of work to get this rebased onto the newer codebase and not really something I can take up again right now - sorry! More than happy for someone else to take this on.

slavab89:
We've merged https://github.com/reactGo/reactGo/pull/937 so this once can be closed.
Thanks @benhjames for the PR and work 😄

PR #619, opened 2016-10-21
Completed logout loop

rmoylan:


slavab89:
@GGAlanSmithee An opinion on this? Seems viable to do this in case we want a dedicated logout screen.
If we this its not mandatory then we should cancel this change.

GGAlanSmithee:
First, apologies to you @rmoylan for not receiving a response in such a long time - thanks for submitting this PR! I am having a hard time seeing what value this brings. I think the standard way is to redirect to the startpage after logging out.

slavab89:
I agree with you. Although it is nice to have a logout page, it seems like the more common use-case is to redirect to the start page like we have now.
For now, lets close this PR.
@rmoylan Thanks a lot for the PR though 😄

PR #648, opened 2016-11-01
Bugfix local auth

shredd:
Fixed local auth. See: https://github.com/reactGo/reactGo/issues/546
This bug occurs with the postgreSQL interface


choonkending:
@shredd You champion! Thanks for fixing!

PR #655, opened 2016-11-03
Fetch data / handle multiple services

### LINKED ISSUE #527: Need to handle multiple services
Note: This is just a template, so feel free to use/remove the unnecessary things
## Enhancement

**Reason to enhance/problem with existing solution**
Currently, preRenderMiddleware only can handle one data. It should be possible to handle multiple data as an app scales.

**Suggested enhancement**
Need to change some structure of preRenderMiddleware.js or fetch-data scripts to accept array of services.

StefanWerW:
Currently, preRenderMiddleware only can handle one data. It should be possible to handle multiple data as an app scales. It also have some problems with reducers, discussed in depth on [this issue](https://github.com/reactGo/reactGo/issues/527) 

I'm proposing a refactor based on @choonkending 's [idea (Approach I)](https://github.com/reactGo/reactGo/issues/487#issuecomment-246284845)  and a modified version of @GGAlanSmithee 's [code](https://github.com/reactGo/reactGo/issues/527#issuecomment-250830554)

-------------------------------------

Advantages:

- Can handle multiple services
- `fetchData` can be either falsy (`null`, `undefined` etc), an action (`function`, `Promise`) or an `array` of either
- Reducer handle specific data fetched (ex: topics) and not in a generic way.
- Data fetchers can use params and query for specific page (ex: https://www.youtube.com/results?search_query=react  --> fetch data for query  `search_query:react` )

------------------------------------

What do you think of this PR? Should I add more stuff? Should we go other direction?

--------------------------------

closes #527


app/middlewares/preRenderMiddleware.js
```diff
@@ -1,10 +1,8 @@
...
+import {flatten} from 'lodash'
 
-function preRenderMiddleware({ routes, location, params }) {
-  const matchedRoute = routes[routes.length - 1];
-  const fetchDataHandler = matchedRoute.fetchData || defaultFetchData;
-  return fetchDataHandler(params);
+//Dispatch is from the store
+//FetchData is either falsy (null, undefined etc), an action (function, Promise) or an array of either:
+export default(dispatch, { routes, params, location: { query } }) => {
+  const dataFetchers = flatten(routes.filter(({fetchData}) => !!fetchData).map(e=>e.fetchData))
```
StefanWerW: Yeah, I notice. If you look close I have a modified version of your code.
```Javascript
//Yours original:
const dataFetchers = flatten(routes.filter(({fetchData}) => !!fetchData))
//Mine:
const dataFetchers = flatten(routes.filter(({fetchData}) => !!fetchData).map(e=>e.fetchData))
```

I thought routes were already flatten.

I do agree that verbose is more readable. But do we really need the first `flatten`?

StefanWerW: @GGAlanSmithee Thanks again for your code! 😄

PR #675, opened 2016-11-09
Get betta with meta

choonkending:
# Reason for Change
- 🐛 There is no simple way to change `title` or `meta` tags on route changes
- Related to https://github.com/reactGo/reactGo/issues/457

# High Level Changes
- Create a `Page` component which composes `Helmet` and children you give it. It also accepts `metaData` as props. Eg. below
```javascript
  render() {
    return (
      <Page metaData={this.getMetaData()}>
        <AboutContainer {...this.props} />
      </Page>
    );
  }
```
- Extract page rendering function into `pageRenderer.js`
   - Extract string interpolation of html page into `index.html`
   - Use `html-loader` to load `<meta>` and `<link>` tags
   - Remove `Meta.jsx` - we no longer use `Helmet` for server-side rendering 🔪 
   - Remove `helmconfig` 🔪


webpack/common.config.js
```diff
@@ -0,0 +1,40 @@
+var path = require('path');
```
psimyn: :100: finally!

app/index.html
```diff
@@ -0,0 +1,68 @@
...
+
+    <meta charset="utf-8">
+
+    <!-- Setting IE=edge tells Internet Explorer to use the latest engine to
+      render the page and execute Javascript -->
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+
+    <!-- Meta descriptions are commonly used on search engine result pages to
+      display preview snippets for a given page. -->
+    <meta name="description" content="Your One-Stop solution for a full-stack
```
psimyn: don't we want these values to come from some kind of config?

psimyn:
doesn't this mean we lose meta server render?

not tested, but code changes seem fine :+1:

choonkending:
> doesn't this mean we lose meta server render?

Good point 👍 We still need to apply `meta` changes from the `Page` component to this, so we should still need to rely on Helmet to rewind. Otherwise, the values in `index.html` will always be hardcoded.

choonkending:
Will add:
`Helmet.rewind()` in `pageRenderer` - but probably tonight.

brtw:
Thank you! I really appreciate the work you're putting into this. Saving my life

choonkending:
@psimyn Please have a look! The last 5 commits should fix the issue about server-rendering that you found.

I'm not particularly happy with the separation of how we apply `meta` data changes in the app though. Would be good to have some opinions.

With the implementation I've gone with:
**Hardcoded meta tags**
- We have hardcoded `<meta>` tags that are stored in the `index.html` file. These use `html-loader`'s string interpolation in order to require the proper files.
   - They are non-page specific
   - I still see value in keeping these in the index.html because they are used everywhere.
**Default meta tags that can be overridden**
- I've moved the important `<meta name="description">` tag to `defaultMetaData`, which only gets set if no values are provided on a certain page. We then use `Helmet.rewind()` to get the values to be replaced in `!!META_TAGS!!` in `index.html`.
- I use Helmet to because nested components will override duplicate head changes
**Link tags + script tags**
- Generated in `pageRenderer.js` 
- Does not rely on react-helmet
- As a side node, I am hoping to move towards non-hardcoded static assets for scripts and links using plugins such as [webpack-manifest-plugin](https://github.com/danethurber/webpack-manifest-plugin) for caching.

The number of places where a `meta`, `link` and `script` tags can be mutated seems to be more than necessary.

**Options**
1. Move all configurations for meta, link tags from `pageRenderer` and `index.html` back into a configuration file which are used by `react-helmet`
 
*Pros*
- All configuration in one place
- Less confusing

2. Keep as-is.
- Hardcoded app level config in `index.html`
- Overridable meta tag changes in `defaultMetaData`
- Link and script changes in `pageRenderer`.

3. Move `link` and `script` changes to configuration file (`defaultMetaData` -> `headAssetsConfig`)
- Keep app level config in `index.html`
- `Helmet.rewind()` will output meta tags
- `pageRenderer.js` will append script and link tags based on headAssetsConfig.

Thoughts?

app/index.html
```diff
@@ -0,0 +1,65 @@
...
+    <meta name="mobile-web-app-capable" content="yes">
+    <meta name="application-name" content="reactGo">
+    <link rel="icon" sizes="192x192"
+    href="${require(`images/chrome-ninja192-precomposed.png`)}" />
+
+    <!-- Add to homescreen for Safari on iOS -->
+    <meta name="apple-mobile-web-app-capable" content="yes">
+    <meta name="apple-mobile-web-app-status-bar-style" content="black">
+    <meta name="apple-mobile-web-app-title" content="reactGo">
+    <link rel="apple-touch-icon" href="${require(`images/apple-ninja152-precomposed.png`)}">
```
psimyn: same deal for icons. could either standardise the names or get out of config?

psimyn:
+1 for Option 3 - I think most of the stuff in <head> can come from config file. Then title/desc and other page-specific stuff can be overridden per-route.

Some other CLI/tool projects have a top-level app.config with app name, icons, description etc.

choonkending:
@psimyn Good to have you on board this 🚢 

I agree. My main concern from moving all these configurations to a config was because of this line https://github.com/reactGo/reactGo/pull/675/files#diff-d352033be8a8816fab63bb99c6e9f47bR24. It was an either `use all default config` or `use config from page` approach.

Most of the comments above kind of reveal the problems with the above approach. We don't want to override base variables, we want to merge them. 

I created a Base level `<Page>` component which had 1 `<Helmet>` element. And every nested route used this `Page` component. This isn't leveraging react-helmet's [features](https://github.com/nfl/react-helmet#features). 

What I'm going to do now:
- Move all `meta`, `icon`, `link` from `index.html` to a base config - let's call it `headAssetsConfig` - this is similar to `helm-config` from before, but will reside in `/pages` directory
- Create a base level <Page> which loads the base config from `headAssetsConfig`. We will probably use `<Page>` for `App.jsx`.
- All nested routes use `<Page>` as well, but Helmet will handle duplicate tags and merge the config correctly.

Hopefully this will clean everything up. Sorry about the huge changes in approaches, but I'm trying to ensure whoever uses this can have a clear understanding of where/how to make changes for SEO without too much of an overhead.

package.json
```diff
@@ -84,6 +84,7 @@
     "extract-text-webpack-plugin": "^1.0.1",
     "file-loader": "^0.9.0",
     "helmet": "^2.1.2",
+    "html-loader": "^0.4.4",
```
choonkending: Given that we're no longer interpolating `require`, and pushing all the configurations into `headAssets`.

It seems like we will no longer need `html-loader`. 

We could instead write a function 

```javascript
const createTemplate = ({ meta, link, title, script }) => (
`
<html>
   <head>
      ${title}
      ${meta}
      ${link}
   </head>
</html>
`
)
```

Unless you can see a benefit in keeping html-loader?

app/index.html
choonkending: Done!

choonkending:
Going to merge in 1 hour 🚢  🚀

package.json
choonkending: I think it still makes sense to separate the `index.html` to it's own file. We can remove `html-loader` in a separate PR if needed (unless anyone objects before merge time)

ZeroCho: The less dependencies, the better! Thanks @choonkending

choonkending: @ZeroCho has spoken! True to our aim of going trimming this repo! Alright! Will remove it in the next upcoming commit.

Issue #709, opened 2016-11-17
Problem when implementing async routing


ZeroCho:
### Description
- Type: Bug

---------------------------------------------------------------
## Bug
**Expected Behavior**
When implementing async routing, each route should be bundled separately, so that each page can be loaded separately when I visit that page.

**Actual Behavior**
Whole pages are served at once. So initial loading is very slow. I found out that this problem is due to `pages/App`. We don't need this because we already have `containers/App`, and each pages can still access `Helmet` without `pages/App`.
Actually, App is not a stand-alone page but a container
I'm preparing PR for this.

**Steps to Reproduce**
It happens when implementing async routing, so there's no possible way to reproduce my problem.

choonkending:
Thanks for raising this! The entire page is served at once?

I actually implemented the meta changes without a `pages/App` initially, and introduced it later on in https://github.com/reactGo/reactGo/pull/675/commits/d3f7e61ab9b0e25bbf8a59683f11c05840979786 for page level common config.

E.g.

`<meta>` tags that were shared across different pages. So then the nested routes would then not need to require and override/resolve all the additional tags. 

More of the thought process was documented https://github.com/reactGo/reactGo/pull/675#issuecomment-260159972 (it's quite long, sorry!)

I am not quite sure how to solve this yet at the moment though :(

choonkending:
Is it because the client side router becomes slower with the `pages/App` having to deal with app specific meta tags as well?

ZeroCho:
I personally removed `pages/App` in my project. Because all pages share `pages/App`, webpack doesn't recognize each page as separate one, so it bundles everything in one js file, although I used `require.ensure` to bundle it separately.
After I removed `pages/App`, webpack began to recognize pages as separate ones.
For page level common config, I'm now using concat to server metas and links because there is no `pages/App`
```
import { link as linkAssets } from 'config/headAssets';
...
pageLink = () => (linkAssets.concat([
    { rel: 'stylesheet', href: '//cdnjs.cloudflare.com/ajax/libs/prism/1.5.1/themes/prism-okaidia.min.css' },
    { rel: 'stylesheet', href: '//cdnjs.cloudflare.com/ajax/libs/prism/1.5.1/plugins/line-numbers/prism-line-numbers.min.css' },
  ]));
```

choonkending:
Ah I see! I personally haven't played around with separate bundling so far, really have to look into that.

choonkending:
Interesting. 

[...]

PR #710, opened 2016-11-17
Fix for async routing

ZeroCho:
Fix for #709


ZeroCho:
@choonkending here I implemented async routing. If you make `page/App` and test it again. You'll find it will not work.

PR #711, opened 2016-11-17
Do not check linebreak-style for windows

ZeroCho:
Windows = CRLF
Unix/OSX = LF
Current linebreak style = LF
I use windows...


choonkending:
Thanks for pointing that out!

Issue #705, opened 2016-11-17
Cull the unneeded dependencies!


choonkending:
## Enhancement

**Reason to enhance/problem with existing solution**
- The less we support, the simpler this project will become! And the easier it is to optimize and improve on performance and code quality!
- Improves delivery effectiveness! 🚢 

**Suggested enhancement**
- Remove unneeded babel plugins
    - [ ] https://github.com/loganfsmyth/babel-plugin-transform-decorators-legacy
    - [ ] https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-react-constant-elements
    - [ ] https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types
- [x] isomorphic-fetch (we already use axios)

I'm referencing create react app's babel plugins as well, over [here](https://github.com/facebookincubator/create-react-app/blob/master/packages/babel-preset-react-app/package.json)

Please suggest more if you find an unneeded dependency!

jrodl3r:
just leaving "babel-plugin-transform-react-inline-elements"?

choonkending:
You mean remove it?  🎉

jrodl3r:
no. I was just asking..

choonkending:
I might have to double check. It promises to make production code faster but I haven't seen many projects using it.

slavab89:
Some of the dependencies were already removed (linked PR's) so i suggest we close it. We can always try and clean up additional dependencies and open more issues.

PR #732, opened 2016-11-27
Remove isomorphic fetch

choonkending:
Part of #705 

Unneeded dependency!

Aint nobody got time for an unneeded dep

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,7 @@
+2.0.1
+===
+- Remove isomorphic fetch as dependency - no longer needed! Hooray!
+
 2.0
 ===
 - #730 - Remove promise middleware
--- a/app/actions/topics.js
+++ b/app/actions/topics.js
@@ -6,17 +6,6 @@ import * as types from 'types';
 
 polyfill();
 
-/*
- * Utility function to make AJAX requests using isomorphic fetch.
- * You can also use jquery's $.ajax({}) if you do not want to use the
- * /fetch API.
- * Note: this function relies on an external variable `API_ENDPOINT`
- *        and isn't a pure function
- * @param Object Data you wish to pass to the server
- * @param String HTTP method, e.g. post, get, put, delete
- * @param String endpoint
- * @return Promise
- */
 export function makeTopicRequest(method, id, data, api = '/topic') {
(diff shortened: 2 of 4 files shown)


choonkending:
Self LGTM! Merging!

lapinskicho:
Do we still need es6-promise?  Is it for providing polyfill for isomorphic-fetch its sole purpose being there? Or it has some other potential use?

choonkending:
Good question @lapinskicho and having laser focus on minimizing dependencies! I'll have a look at lunch to see if `axios` needs a polyfill - otherwise we're removing it!!!

choonkending:
Yep - axios expects the native es6 promise to be polyfilled - which makes sense. Read more [here](https://github.com/mzabriskie/axios#promises)

lapinskicho:
Got it.  Thanks!

For minimizing dependencies, I found two more modules that may not be in (actual) use, namely "express-flash" and "method-override" - or please let me know if I missed something in the code that they are actually needed.

lapinskicho:
Also now in the code, each of ./app/actions/users.js and ./app/actions/topics.js are calling polyfill() for axios, but ./app/services/topics.js is not calling polyfill() (a bug?) - seems like we should have a utility file that takes care of the polyfill-ed axios and export for all other modules to use?

ZeroCho:
Method override is for future use of http PUT or DELETE request.
And I'm also wondering why polyfill() is needed because it can be put inside webpack.config.js

lapinskicho:
My first thought was about server-side, so I checked Node 5 (this package requirement) support of Promise, and it doesn't support all of the es6 promise implementation.

http://node.green/#Promise

choonkending:
>  each of ./app/actions/users.js and ./app/actions/topics.js are calling polyfill() for axios, but ./app/services/topics.js is not calling polyfill() (a bug?) - seems like we should have a utility file that takes care of the polyfill-ed axios and export for all other modules to use?

Yes you are right @lapinskicho ! Thanks for finding that. I think we should only need to call the polyfill once to be honest, possibly in the main file. 

@ZeroCho Do you have a webpack plugin for a polyfill? 

I'm creating an issue for this haha - so we can continue our conversation over there peeps!

PR #780, opened 2016-12-20
ESLint fix

ZeroCho:
There are too many errors when linted by ESLint.
I fixed some of them without any opinionated things. (I just followed ESLint instruction, spacing, unused-variable, relative path for module, etc.)

There are some more errors to be fixed, but it needs sharing opinions to fix those, so I left them untouched.

@choonkending @psimyn Please review this!

### DIFF
--- a/app/services/index.js
+++ b/app/services/index.js
@@ -1,2 +1,3 @@
-export { default as voteService } from './topics';
+import voteService from './topics';
 
+export default { voteService };
--- a/.eslintrc
+++ b/.eslintrc
@@ -13,7 +13,7 @@
             "jsx": true,
             "es6": true,
             "experimentalObjectRestSpread" : true
-        },
+        }
     },
     "rules": {
         "comma-dangle": 0,
--- a/app/actions/messages.js
+++ b/app/actions/messages.js
@@ -1,6 +1,8 @@
 /* eslint consistent-return: 0, no-else-return: 0*/
-import * as types from 'types';
+import * as types from '../types';
 
 export function dismissMessage() {
   return { type: types.DISMISS_MESSAGE };
 }
+
+export default { dismissMessage };
--- a/app/actions/topics.js
+++ b/app/actions/topics.js
@@ -2,7 +2,7 @@
 import { polyfill } from 'es6-promise';
 import request from 'axios';
 import md5 from 'spark-md5';
-import * as types from 'types';
+import * as types from '../types';
 
 polyfill();
 
--- a/app/actions/users.js
+++ b/app/actions/users.js
@@ -2,7 +2,7 @@ import { polyfill } from 'es6-promise';
 import request from 'axios';
 import { push } from 'react-router-redux';
 
-import * as types from 'types';
+import * as types from '../types';
 
 polyfill();
 
--- a/app/client.jsx
+++ b/app/client.jsx
@@ -3,10 +3,10 @@ import { render } from 'react-dom';
 import { Provider } from 'react-redux';
 import { Router, browserHistory } from 'react-router';
 import { syncHistoryWithStore } from 'react-router-redux';
-import createRoutes from 'routes';
-import * as types from 'types';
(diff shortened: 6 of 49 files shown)


ZeroCho:
And about the relative path (for example, css -> ../css, types -> ../types), I changed it into relative path because there are npm modules with the same names such as types, services, reducers, stores.

app/services/index.js
```diff
@@ -1,2 +1,3 @@
-export { default as voteService } from './topics';
+import voteService from './topics';
```
choonkending: I actually like having an explicit `export { default as x } from a` statement, and I've seen it in a couple of places. 

If this is part of the eslint config, perhaps we can remove that rule

choonkending:
@ZeroCho nice one! For the relative paths I'm not too fussed about it given that there might be conflicts with npm modules (but it's weird that there are modules called services and reducers haha). I'd like us to move towards keeping reducers.js , service.js, component, container, test and styles within its own component directory. Something that @psimyn spiked out a while ago,  that approach encapsulates our component and only exposes what we need, so relative paths won't be too much of a problem

app/services/index.js
ZeroCho: `default as voteService` is not a problem, but using at least one export default is recommended in ESLint. [Link](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/prefer-default-export.md)
And for scalability, when other services are added, we need `export default` that exports whole services at once.

choonkending: Good point - it makes sense to have one default export. I think we can keep that eslint rule for now. I was hoping we could do something along the lines of https://github.com/reactjs/redux/blob/master/src/index.js#L29...L35 which looks pretty sick hehe.

choonkending:
Thanks @ZeroCho ! We might move towards a component directory such as
```
example-component/
   - index
   - View
   - Container
   - reducer
   - service
   - test/
   - style.css
```

in the future - but I'll put issue into the project!

choonkending:
Awesome effort fixing the eslint! 🎉

app/services/index.js
choonkending: I think this bit caused #784 . Reading [ponyfoo's es6 article](https://ponyfoo.com/articles/es6), [this](https://ponyfoo.com/articles/es6-modules-in-depth#exporting-a-default-binding) and [this](https://medium.com/@kentcdodds/misunderstanding-es6-modules-upgrading-babel-tears-and-a-solution-ad2d5ab93ce0#.chotb1xzh) article. It seems I might have misunderstood ES6 imports and exports

@ZeroCho For now, I think we should either:
1. Revert to what it was before (for this file)
2. `export {}` alongside with `export default {}`. This makes consuming this module confusing as there are several ways to use this module.
3. `import defaultModule` instead of `import { foo } from './bar'`.

```js
import service from '../services';
```

I'm still thinking which is the best approach

ZeroCho: @choonkending Okay, I'll test it again!

choonkending: Reading Axel's [article](http://www.2ality.com/2014/09/es6-modules-final.html) about ES6 module syntax

> The default export is just another named export

I think for this reason, when we `export default expression`, we cannot immediately `import { foo } from 'module'` as `default` export is just another named export.

Personally, I like de-structuring in `import` statements as I find it easy to comprehend. 

Although `export default` does seem to be the recommended approach in [airbnb eslint](https://github.com/airbnb/javascript), I think it does rely on our design decisions. 

I think we should use `export default` when we have a module that does 1 thing, e.g. a Component, or a single utility function. In service, the intent behind it was to provide an interface to different service endpoints, I view `voteService` as a single utility function, exposed via the service interface.

What I'm getting is, I think in this case, it is alright to 
```js
export { default as voteService } from './topics';
```

without having a `export default`. I think having both a `export {}` and a `export default {}` might be more confusing to a user looking at this module and duplicates the code.

@ZeroCho What are your thoughts?

Issue #784, opened 2016-12-27
Bug: Todos not persisting in mongo DB


gauravarora91:
git clone
npm install 
npm run dev

> react-webpack-node@2.1.0 dev /Users/**/reactGo
> cross-env NODE_ENV=development nodemon

[nodemon] 1.11.0
[nodemon] to restart at any time, enter `rs`
[nodemon] watching: /Users/**/reactGo/app/utils/**/* app/server.jsx app/routes.jsx /Users/**/reactGo/server/**/* webpack/webpack.config.dev-client.js webpack/webpack.config.dev-server.js .stylelintrc
[nodemon] starting `npm run build:dev && node compiled/server.dev.js`

> react-webpack-node@2.1.0 build:dev /Users/**/reactGo
> cross-env NODE_ENV=development webpack ---debug --colors --display-error-details --config ./webpack/webpack.config.dev-server.js

Hash: 401b2505459a9dbe1331
Version: webpack 1.14.0
Time: 11492ms
            Asset     Size  Chunks             Chunk Names
    server.dev.js   876 kB       0  [emitted]  server
server.dev.js.map  1.01 MB       0  [emitted]  server
    + 309 hidden modules
--------------------------
===> 😊  Starting Server . . .
===>  Environment: development
===>  Listening on port: 3000
===>  Using DB TYPE: MONGO
--------------------------
===>  Succeeded in connecting to mongodb://localhost/ReactWebpackNode
webpack building...
webpack built 32af108534092747fe2b in 9423ms

Problem is mongo has todos still its not showing todo's, on first render as well as via routing. Nothing comes in state as well.

choonkending:
@gauravarora91 Thanks for reporting. I can confirm that this is happening to me as well.

choonkending:
Documenting our discussion here https://github.com/reactGo/reactGo/pull/780/files#r94002355

gauravarora91:
Thanks!

PR #786, opened 2016-12-28
Fix failed imports with named export

choonkending:
## Why
#784

## High Level Changes
Removed default export as we see fetch-data and services as utility modules that
expose distinct functions, not necessarily grouped and imported together.

### DIFF
--- a/app/fetch-data/index.js
+++ b/app/fetch-data/index.js
@@ -1,4 +1,2 @@
-import { fetchVoteData } from './fetchVoteData';
-
-export default { fetchVoteData };
+export { default as fetchVoteData } from './fetchVoteData';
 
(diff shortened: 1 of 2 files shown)


choonkending:
@ZeroCho although it seems to be not following best practice, imo it seems more like a design decision. Let me know what you think before I 🚢

ZeroCho:
@choonkending I read your article, and I agree it's just design problem(opinion-based).
Just one more question before merging it. If a user created another service such as userService, is it okay to code like below?
```
export { default as voteService } from './topics';
export { default as userService } from './users';
```
If it's okay, everything is fine. Our way is scalable.

choonkending:
Yep that's totally fine! The syntax says "export the default import from topics as { voteService".

Initially inspired by https://github.com/gaearon/react-dnd/blob/master/src/index.js#L1 and found it to be pretty simple to read!

choonkending:
Thanks!

PR #789, opened 2016-12-29
Remove inline environment variables webpack plugin

choonkending:
- Part of #705 

## High Level Changes
- Replace [inline-environment-variables-webpack-plugin](https://github.com/tikotzky/inline-environment-variables-webpack-plugin) with [webpack.EnvironmentPlugin](https://github.com/webpack/docs/wiki/list-of-plugins#environmentplugin)

**Background**
- Was introduced in https://github.com/reactGo/reactGo/issues/157 but with this change we can remove an extra dependency

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,10 @@
+2.1.1
+===
+- PR #786 - Fixed failed imports with named export
+- PR #780 - Eslint fix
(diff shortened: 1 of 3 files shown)


choonkending:
Self 👊

Issue #742, opened 2016-11-30
Centralize Configurations


choonkending:
# Why

- We have configurations files within:
    - `app/config/app.js`
    - `server/config/appConfig.js`

My proposal is to move all of these to `config/app.js` which will contain all the files. Eliminating confusion.

ZeroCho:
Good idea. I hate duplication, too.

choonkending:
Update of what I've done so far:

https://github.com/reactGo/reactGo/compare/master...choonkending:centralize-config

- Move `app/config/app` to `config/`
- Move `express`, `routes`, `passport` in `server/config` out to individual files
- Move remaining `server/config/*` to `config/`

Still need to merge `config/app` and `config/appConfig`

choonkending:
Having a bit of trouble `babel`-ifying some of the directories outside of `server` in production.

I think we should run webpack through `server/index.js` instead of `server.jsx`.

Basically http://jlongster.com/Backend-Apps-with-Webpack--Part-I

choonkending:
Well this exercise is going to take me a while

choonkending:
Time to shave some yaks 

![image](https://cloud.githubusercontent.com/assets/3787885/20780354/4aeb1230-b7ce-11e6-87eb-342d305142df.png)

choonkending:
I'm gonna do #746  first because that's what you do when you shave a yak

slavab89:
What i actually did in my project was the following:
I used ```node-config``` as my configuration manager because i had several environments and wanted the ability to control all of them in 1 place (both server and client side).
So i've create a folder config at the root directory (like you did) and places all the config files there.
In the server side (the express app) i can just use those files and the config module as usual because, well its there.
On the client side its a little more tricky. Because my project is using a remote API server, i had to pass the configuration down to the browser side of the app.
For this, once i loaded the config, i placed it at the express.appLocals variable so that every request that comes to the server would have this config available to it.
After that, when creating the redux store, i added this configuration to the store and then i was able to access it from the client side as well (cool =] )
To sum up:

- Used ```node-config``` module to load the configuration for the project
- Once ```express``` loads up, i put the client configuration on the ```app.locals``` variable that express exposes
- Once a request arrives to the server i bootstrap the ```redux store``` with that config from the ```app.locals``` of express
- Access the config from the ```redux store``` where needed (for example to get the URL of the remote API server)

choonkending:
[...]

PR #792, opened 2017-01-02
Centralize config/

choonkending:
# Why
Part of #742 

Having `config/` folders in `app` and `server` with shared functionality makes it **harder to understand** and maintain our app. 

With this change, we move our `config/` to the root level. I have attempted to logically the configurations into separate files.

# High Level Changes
- Move `config/headAssets` -> `pages/assets`
- Centralise `config/` at the root level (`/`)
   - `app/config/app` -> `config/app`
   - Add `config/env` to contain `process.env` variables
   - Move `server/config/secrets` to `config/secrets`
   - Move and rename `server/config/constants` to `config/dbTypes`
   - Move and merge variables in `server/config/appConfig` into `config/env`
- Use `baseURL` in `server.jsx` instead of constructing the baseURL from host and port
- Webpack Changes
    - Remove `__DEVCLIENT__` and `__DEVSERVER__` defined in `DefinePlugin`. Instead use `EnvironmentPlugin` to receive `NODE_ENV` variables. We reduce the dependency on `webpack` to define these variables and utilise a more straight-forward way of defining `production` and `development` modes

## Remaining changes
- Cleaning up `server/config` folder
The remaining files in this folder are not really an app level concern to me. They are still configurations, but I think we can rename this folder to be more **intention-revealing** rather than a `config`, which could be anything. Perhaps something like `boostrap/`? ❓ 

`config` files to me indicate that they are **values** rather than functions. These values are used app wide to determine what "mode" or "function" to use. This is partially why I think the `server/config` files should be renamed to something more imperative, as they are functions.

## How to review
This is a pretty huge change, but I've separated them into logical commits. Hopefully as you review you can step through each commit to see the incremental changes I've made.

@ZeroCho @psimyn I'd like you to have a look! (Or whoever is interested + have the context)


choonkending:
@psimyn has given it the LGTM! Merging by noon!

Issue #796, opened 2017-01-06
A typo in common.config.js ('.bit' should be '.bin')?


lapinskicho:
webpack/common.config.js:8:    return ['.bit'].indexOf(x) === -1;

Is it meant to exclude the node_modules/.bin directory?

choonkending:
Good find! Thanks for raising this issue!

choonkending:
Thanks @lapinskicho ! Fixed!

PR #798, opened 2017-01-07
Exclude binary files correctly for webpack

choonkending:
Raised in #796 

We were excluding binaries incorrectly. Thanks to @lapinskicho for spotting this!


choonkending:
Straight forward change. Merging!

Issue #794, opened 2017-01-04
config and server/config might be confusing


ZeroCho:
After centralizing config #792 , there is still a config folder in server directory.
I think we can simplify that by making folder structure like below.
```
--server(folder)
----db(folder)
----passportStrategy(folder)
------local.js
------google.js
----express.js
----passport.js
----routes.js
----index.js
```
or maybe just changing config folder name into another.

GGAlanSmithee:
This makes a lot of sense. I remember being confused by this when first using this project. especially since a lot of the code in that folder is not config, but express middleware / routes, db models etc.

ZeroCho:
@GGAlanSmithee @choonkending I think just renaming folder is better.
```
--server(folder)
----index.js
----db(folder)
----???(folder, previously config)
```
I have bad naming skills, especially in english. So can you guys rename it?

choonkending:
You peeps are the best! 

When I did https://github.com/reactGo/reactGo/pull/792#issue-198292090 I decided to keep the remaining `server/config` files out because I couldn't decide whether or not to move them out or rename it.

>  especially since a lot of the code in that folder is not config, but express middleware / routes, db models etc

Exactly!

What is an intention-revealing name for a folder? I agree with @ZeroCho that renaming the folder would make sense. 

I was thinking:

- bootstrap
- init

Any other ideas? This was my next PR planned for clean up, but I still haven't decided which name to use.

GGAlanSmithee:
IMO, it would be enough to remove the `config` folder, moving it's content one level up. (and moving /renaming `passport.js` to `passport/index.js`

choonkending:
@GGAlanSmithee That's what I initially attempted as well. 

When I moved the content one level up, the file names became a bit too generic to my liking. I then renamed them to `loadPassport`, `loadRoutes`, `loadExpress` to indicate that they were methods that we called to bootstrap our application.

```js
server/
   - /db
   - index.js
   - loadPassport.js
   - loadRoutes.js
   - loadExpress.js
```

Given that they are all used for loading/bootstrapping/initialising our application, it then made sense to just rename `config` to indicate what those files were actually used for (which is what @ZeroCho mentioned).

For now, I'll rename the folder to `init` and make `passport.js` into `passport/index.js`. If we find that it's confusing still, we can revisit this issue in the future. 

Thanks for all the suggestions!

PR #800, opened 2017-01-07
Rename server/config -> server/init

choonkending:
Raised in #794 

# High Level Changes
- Rename /config -> init in server/
- Move `passport.js` into `passport/index.js`


choonkending:
Self 👊

PR #802, opened 2017-01-07
Simplify npm commands

choonkending:
# Why?

-  To run a prod build, we will always do the following:
```bash
npm run build && npm start
```

Simplifying this a bit, we could just do `npm run build` as part of `npm start`. This also makes the production command consistent with the dev command.

Commands are now:
```bash
// Builds and runs dev server
npm run dev 

// Builds and runs production server
npm start
```

- Remove `postinstall` as no longer needed with the above change
- Removing `keep-green` command as we disabled greenkeeper

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,11 @@
+2.2.4
+===
+- PR #802
+  - Add `npm run build` as part of `npm start`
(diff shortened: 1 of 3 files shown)


choonkending:
Self 👊 as it's a straight forward change!

Issue #757, opened 2016-12-08
How to load multiple data with preRenderMiddleware?


k1tzu:
It looked like this before:
```
static need = [  // eslint-disable-line
        fetchData,
         fetchAnotherData
     ];
```

And now I'm trying to do:
```<IndexRoute component={Dashboard} fetchData={fetchData, fetchAnotherData} />```
and it doesn't work. It returns only "fetchAnotherData" results.

I'm completely stuck, help me please!

**Comment by StefanWerW** [comment]
Have a look on this #655 . I use this modified version of preRenderMiddleware and it does the job.

Advantages:

- Can handle multiple services
- `fetchData` can be either falsy (`null`, `undefined` etc), an action (`function`, `Promise`) or an `array` of either
- Reducer handle specific data fetched (ex: topics) and not in a generic way.
- Data fetchers can use params and query for specific page (ex: https://www.youtube.com/results?search_query=react  --> fetch data for query  `search_query:react` )

**Comment by k1tzu** [comment]
I can't understand something about you code:
server.jsx
```
preRenderMiddleware(store.dispatch, props)
      .then(data => {
                           store.dispatch({ type: types.FETCH_DATA_SUCCESS, data });
```
so we're looking forward here to get some data from the database as the result of preRenderMiddleware? Right?

app/middlewares/preRenderMiddleware.js
```
return Promise.all(flatDataFetchers.map(dataFetcher => dispatch(dataFetcher({ ...params, query }))))
```
It returns only when all fetchers are complete. 

But if we look into the function it calls:
app/fetch-data/fetchVoteData.js
```
const fetchData = () => {
  return {
    type: types.GET_TOPICS,
    promise: voteService.getTopics()
  }
}
```
We'll see that it returns an object with 'type:' and 'promise:' keys. But not the data from the database. Is it normal?
```
      .then(data => { // {type: types.GET_TOPICS, promise: function }
                           store.dispatch({ type: types.FETCH_DATA_SUCCESS, data });
```

**Comment by k1tzu** [comment]
Ok I got it to return data, not the {pending} promises.
```
import {flatten} from 'lodash'

//Dispatch is from the store
//FetchData is either falsy (null, undefined etc), an action (function, Promise) or an array of either:
export default(dispatch, {routes, params, location: {query}}) => {
    // get all routes (that has fetchData) as a flat array
    const flatRoutes = flatten(routes.filter(({fetchData}) => !!fetchData));

    // get each routes fetchData
[...]

PR #806, opened 2017-01-09
Upgraded Webpack to version 2.2.0-rc.3

lapinskicho:
1) Upgraded Webpack to version 2.2.0-rc.3; upgraded dependencies to use this latest Webpack version.

2) Refactored webpack.config.js for Webpack 2.  Also consolidated webpack/\*.config.\*.js files into a single webpack.config.js file.

3) Added postcss.config.js as it is necessary for postcss-loader in Webpack 2.

4) Modified karma.conf.js for Webpack 2.  Also moved karma.conf.js and tests.webpack.js to app/tests/ - may not be necessary but could be easier to fix the relative path lookup errors in the tests - the tests couldn't run before this migration and should be fixed in a different PR.

5) Updated package.json and nodemon.json to reflect the Webpack 2 upgrade.

6) This check-in does not introduce any new eslint errors.

7) Notice that stylelint-webpack-plugin does not support Webpack 2 yet and have been commented out.

8) UNMET DEPENDENCIES ERROR on webpack-2.2.0-rc.3 initiated by extract-text-webpack-plugin 2.0.0-beta.4 can be safely ignored - extract-text-webpack-plugin should have added the correct peerDependencies but it didn't in this beta.4 version.

9) Some typos in webpack/webpack.config.prod.js have been fixed (e.g. webpack.optimize.UglifyJsPlugin takes "compress" as option key, not "compressor")

NOTE: engines.node is upgraded to ">=6.x" in package.json - which may not be necessary; but I am using node version 6.9.1 on my local development environment, so I don't have node 5.x to test.


lapinskicho:
Sorry found a bug:

app/tests/karma.conf.js

line 56: exclude: path.resolve(__BASE_DIR__, '/node_modules/')
should be
line 56: exclude: path.resolve(__BASE_DIR__, 'node_modules')

choonkending:
@lapinskicho you champion! And thanks for the high-level details which makes it easier for me to review. Since @ZeroCho has experience with the migration, I'm pulling him in for review as well.

From an initial skim, code looks really good and clean. I would suggest breaking the giant `webpack` config into several smaller pieces under a `/webpack` folder still, just to manage the complexity, but I really like how you've simplified it. But I will need a deeper dive into the configurations in order to review this better. 

Are there any reference implementations/recommended readings for migrating you based off your branch (just so I can read and catch up while reviewing?)

ZeroCho:
Wow. I learned a lot from his code.
Im using stylelint-webpack-plugin with webpack@2, but I'll check it again.

server/index.js
```diff
@@ -7,6 +7,10 @@ import initExpress from './init/express';
 import initRoutes from './init/routes';
 import renderMiddleware from '../app/server';
 
+const webpackDevMiddleware = require('webpack-dev-middleware');
```
GGAlanSmithee: maybe require these conditionally inside the `isDebug` if statement, since they're only used there?

lapinskicho:
@choonkending I find it easier to manage and maintain with everything in one file.  But please refactor the code the way you see fits to the project.  I will follow up with some links for Webpack 2.

@ZeroCho yes, my bad. I just looked up stylelint-webpack-plugin 0.5.1 and it only warns about the UNMET DEPENDENCIES ERROR on webpack-2.2.0-rc.3 but it runs without errors (like extract-text-webpack-plugin).

@GGAlanSmithee http://eslint.org/docs/rules/global-require - there are pros and cons putting require elsewhere in the code.  I have eslint rule global-require turn on my projects, but I just checked reactGo has it turn off.  So I guess it has been agreed that way?  @choonkending please make any appropriate changes.

lapinskicho:
@choonkending http://javascriptplayground.com/blog/2016/10/moving-to-webpack-2/ this is an overview how to migrate from Webpack 1 to 2.  I tried to keep the config "as is" - so I didn't apply some of the optimization suggested in the article, such as I kept webpack.optimize.OccurrenceOrderPlugin() in the code, and didn't stop Babel from compiling ES2015 modules.  

For postcss-loader, https://github.com/postcss/postcss-loader has Webpack 2 covered.

After all, it is hard to test if webpack output works 100%.  We just "assume" it works when we see things we expect to see...

GGAlanSmithee:
@lapinskicho agreed that it is stylistically worse, but since that code is only used dev mode, there is an actual difference to include those files unconditionally. Since that code is `require`d rather than `import`ed, webpack cant tree shake it either AFAIK. If it is to be kept in the top scope, it would make more sense to `import` it, if possible. Probably won't matter much in reality though, so I guess it's not a huge issue.

lapinskicho:
@GGAlanSmithee Thanks for pointing it out.  I thought Webpack would eliminate dead code (the if-block process.env.NODE_ENV === "development" when it equals to "production") but I just looked at compiled/server.js it is still there.  Maybe it is the job of Uglify webpack plugin.  Anyway, you are correct to stick with es6 import/export in the top scope.

lapinskicho:
https://github.com/webpack/webpack/issues/2867

Turns out Webpack 2 tree shaking feature is broken.  That's why...

server/index.js
ZeroCho: I also turn off global-require, and put it in `isDebug`

GGAlanSmithee:
@lapinskicho 

> Turns out Webpack 2 tree shaking feature is broken. That's why...

Yeah, it has been shaky (pun intended), but I am still pretty sure it wil never be able to treeshake `require` statements

server/init/express.js
```diff
@@ -25,7 +25,7 @@ export default (app) => {
   app.use(bodyParser.urlencoded({ extended: true })); // for parsing application/x-www-form-urlencoded
   app.use(methodOverride());
 
-  app.use(express.static(path.join(__dirname, '..', 'public')));
+  app.use(express.static(path.join(process.cwd(), 'public')));
```
choonkending: Nice one. This is much cleaner and feels more correct!

webpack.config.js
```diff
@@ -0,0 +1,234 @@
...
+      ];
+    }
+    if (!production && browser) {
+      return [
+        new webpack.HotModuleReplacementPlugin(),
+        new webpack.NoErrorsPlugin(),
+        new webpack.EnvironmentPlugin(['NODE_ENV'])
+        // stylelint-webpack-plugin does not support webpack 2 yet
+        // new styleLintPlugin({
+        //   configFile: path.resolve(__DIR__, '.stylelintrc'),
```
choonkending: Has anyone found stylelint-webpack-plugin useful for us? Or should we just 🔪 it?

webpack.config.js
```diff
@@ -0,0 +1,234 @@
...
+        new webpack.BannerPlugin(bannerOptions)
+      ];
+    }
+    if (production && browser) {
+      const filename = 'styles/main.css';
+      const allChunks = true;
+
+      return [
+        new ExtractTextPlugin({ filename, allChunks }), // extracted css to css/main.css
+        // new webpack.optimize.CommonsChunkPlugin({ names: ['vendor', 'manifest'] }),
```
choonkending: I was meaning to extract `vendor` and `manifest` from our bundle for caching, but we should probably just remove this comment for now

choonkending:
Nice one. I'll pull your branch and attempt to see if I can break the webpack file into smaller bits and see if it logically fits together when I get the chance tomorrow (or this week, I'm still attempting to clean up #757 ). Really happy with how much you've improved the webpack config. 👍

choonkending:
Just submitted an initial PR to your branch @lapinskicho 

Additional changes I'd like to make to your PR later on
- Attempt to conditionally require the development config on dev mode. Linting is a broken currently. https://github.com/reactGo/reactGo/pull/806#issuecomment-271473590
-  Split out the file into smaller modules.
```
/webpack
    /loaders
       - cssLoader
       - babelLoader
       - etc
    /plugins
```

Although I agree with you with this

> I find it easier to manage and maintain with everything in one file

I am aiming to break them down and but name them appropriately so that they are not hard to follow. Reducing the complexity of this file will make it easier to maintain later on. 

You've done a great job with fixing so many of the issues I found with our initial webpack configuration! 👍 👍

webpack.config.js
```diff
@@ -0,0 +1,234 @@
...
+    const serverSide = 'css-loader/locals' + queryStr + '!postcss-loader';
+    const browserSide = 'style-loader!css-loader' + queryStr + '!postcss-loader';
+
+    const myLoader = browser ? browserSide : serverSide;
+
+    const prodBrowserRenderLoader = ExtractTextPlugin.extract({
+      fallbackLoader: 'style-loader',
+      loader: 'css-loader' + queryStrExtractTextPlugin + '!postcss-loader'
+    });
+    const loader = production && browser ? prodBrowserRenderLoader : myLoader;
```
zamiang: Implemented this on a repo based on reactGo over the weekend. One issue I had was that in production the server rendered the dev css module classnames (ie: '.class-name_3efase3') but the css asset was exported using the prodBrowserRenderLoader and rendered the hashed classnames (ie:'.3efase3') so they did not match. Easy way to repo is to run in prod mode and turn off javascript in your browser.

The client would re-render and visually correct it almost instantly (much like it does in dev) but I would expect the server to render the correct classnames.

tl:dr
This should likely be `const loader = production ? prodBrowserRenderLoader : myLoader;`

choonkending: Thanks for that - I'll take a look at it when I submit more PRs to the branch!

choonkending:
Closing in favour of this https://github.com/reactGo/reactGo/pull/822. #822 is a branch based off this one!

Issue #739, opened 2016-11-29
Es6 Promise polyfill 


choonkending:
Continuing the discussion in https://github.com/reactGo/reactGo/pull/732#issuecomment-263272260

# Type: Bug

- We are calling `polyfill` multiple times and inconsistently
- Should we stick we the current polyfill or use a webpack plugin?
   - I'm leaning with sticking with a polyfill dependency because I'm afraid webpack is doing wayyyy too much (but I'm not entirely sure)

@ZeroCho @lapinskicho

ZeroCho:
```
entry: {
    app: ['babel-polyfill', 'eventsource-polyfill', './client', hotMiddlewareScript],
  },
```
this is my webpack.config.dev-client.js. and **not** for server.
eventsource-polyfill is for IE

lapinskicho:
@choonkending I am new to the package and still picking up the nuts and bolts of it, but I think it is good to use babel-polyfill to emulate a full es6 environment (over es6-promise to just emulate es6 promise) as down the road, developers may use additional es6 features (e.g. generators), so it is more for "forward-compatibility."

yowakita:
@choonkending I created a PR to add `babel-polyfill` and `eventsource-polyfill` to the Webpack configurations.
#808

PR #808, opened 2017-01-10
Adds babel-polyfill and eventsource-polyfill to client and production webpack configs

yowakita:
Adds `babel-polyfill` and `eventsource-polyfill` to client and production Webpack configs so that this boiler works on IE.

See #739 & #732


ZeroCho:
This is good but I think this will conflict with #806

ZeroCho:
I'm currently using babel-runtime instead of babel-polyfill in dev environment

yowakita:
Im curious of what advantages babel-runtime have @ZeroCho . You would need to require it wherever the polyfill is needed, correct?

Issue #809, opened 2017-01-10
Rename `server.jsx` to `renderMiddleware` or `applicationMiddleware`


choonkending:
`server.jsx` is confusing as a name. It is executed on the server, but let's name it for what it is.

It is an express middleware that passes the `request` to `react-router` for matching. We then fetch data for the route and render.

Any ideas?

choonkending:
Thoughts so far:
- renderingMiddleware
- expressMiddleware (might be too vague)
- applicationMiddleware

ZeroCho:
+1 on renderingMiddleware because it is for server side rendering.
or maybe SSRMiddleware?

GGAlanSmithee:
Another +1 on renderingMiddleware

@ZeroCho since this file would probably be moved to the server folder (i assume?), it's already implied that it's server-executed.

A small nitpick: I would suggest naming it `renderMiddleware` instead of `renderingMiddleware`, just as you would have `authenticationMiddleware` or `authenticateMiddleware`, and not `authenticatingMiddleware`. This is probably pretty subjective thoguh :)

choonkending:
@GGAlanSmithee Agreed. The only reason why i thought `rendering` was okay was because we previously had `preRenderMiddleware` which didn't make as much sense. I think in this case, `renderMiddleware` is a good name.

As for moving to `/server`, yes I think it will be moved there. There might be other files moved there as well, such as `pageRenderer`, but `fetchDataForRoute` will still be kept in the `app`. 

I'll work on this tonight!

PR #811, opened 2017-01-12
Rename serverjsx -> render/middleware

choonkending:
Part of #809 

# High level Changes
  - Move and rename app/server.jsx -> server/render/middleware.js
  - Move createScripts -> server/render/
  - Move pageRenderer -> server/render/


choonkending:
Self 👊

PR #815, opened 2017-01-25
Use express server port from config file.

mstorm:
Declared server port from `config/env.js` file.
But, It doesn't work.


ZeroCho:
@mstorm You mean this doesn't work? Then why did you make Pull request?

mstorm:
This PR is fixed.

Issue #743, opened 2016-11-30
Spike Webpack 2.0


choonkending:
https://webpack.js.org/guides/

ZeroCho:
Is this stable or under development?

choonkending:
I think they are still working on the docs, and there are still a couple of issues they need to iron out. Maybe I should change the  title to `spike` hehe

psimyn:
iirc docs were the only thing holding back 2.0 release - but other plugins/packages still have some teething issues. API stable enough to making the changes though, I think any dealbreaker issues should be pretty apparent.

sys13:
Webpack 2.2.0-rc.0 is out now
https://medium.com/webpack/webpack-2-2-the-release-candidate-2e614d05d75f#.s6epb3451

lapinskicho:
I have successfully migrated my app based on reactGo to Webpack 2.  As always for Webpack, when it works, it seems to be straightforwarded, but it takes time to tweak the settings to make things work. In general, a few string replacement in the webpack config, "resolve" takes different options as webpack 1, a few modules need to be upgraded to work with webpack 2.2.0-rc.3, and stylelint-webpack-plugin is not compatible with Webpack 2 yet.  Also postcss-loader needs to have its config in a separate file postcss.config.js in the project directory.

```
/**
 * webpack.config.js
 *
 * process.env.NODE_ENV is used to determine to return production config or not (an array with both browser and server config)
 * if not, env is used to determine to return browser-rendering config (for hot module replacement) or server-side rendering config (for node)
 * env is a string passed by "webpack --env" on command line or calling this function directly
 * if env contains substring 'browser', then returns browser-rendering config, otherwise server-rendering config
 * NOTE: browser/server is client/server-side rendering respectively in universial/isomophoric javascript
 */
const fs = require('fs');
const webpack = require('webpack');
const path = require('path');
const ExtractTextPlugin = require('extract-text-webpack-plugin');

module.exports = (env = '') => {
  const isProd = process.env.NODE_ENV === 'production';
  const browserRender = (env.indexOf('browser') >= 0);
  debug.info(`running webpack.config.js; isProd=${isProd}; browserRender=${browserRender};`);

  /**
   * __dirname is unreliable after webpack-ed to another directory
   * so process.cwd() is used instead to determine the correct base directory
   */
  const __DIR__ = process.cwd();

  const PATHS = {
    app: path.resolve(__DIR__, 'app'),
[...]

PR #822, opened 2017-02-05
Webpack 2 is here!

choonkending:
# Reason
#743 - spike webpack 2.0

but mainly due to @lapinskicho 's https://github.com/reactGo/reactGo/pull/806 (thanks again!!)

# High Level Changes
- Use #806 's changes
- Remove stylelint as dependency
- Replace NoErrorsPlugin with NoEmitOnErrorsPlugin
♻️ 
⚠️ The changes here are opinionated, i.e. there are good reasons to keep configurations within a single file. However, I believe that splitting a large configuration file into smaller manageable pieces will be worth its value in the long run as it will make webpack more manageable. I think this outweighs the cons (having extra mental overhead in finding modules). I spent quite some time refactoring and coming up with meaningful names + folder structure, I _really_ hope this will make things more maintainable!

- Move webpack configurations into `/webpack`
- Create `webpack/rules/` which define
    - javascript (babel-loader) - I thought hard about this one. Babel is an implementation detail rather than _what_ we want, which is a loader for javascript with support for jsx + es2015.
    - image (url-loader
    - css (css-loader + postcss-loader + friends)
- Split out plugins to `webpack/plugins`
- Split out externals to `webpack/externals`
- Extract PATHS to `webpack/paths`

**Important upgrades**
- webpack - 2.2.1 🎉 🎉 

🔧 
- Fix issue raised in https://github.com/reactGo/reactGo/pull/806#discussion_r96241135 regarding production server render

# Questions
1. This is a huge change. How do I review it?
Good question. Go through each logical commit. Each commit should be a small change enough to review in nice bite-sized chunks.

Thank you!


choonkending:
> I removed --debug from "build:dev" in package.json, if you want to keep it you may want to add it back.

Thanks for that! I'll add that back in for dev mode!

choonkending:
Merging in an hour!

app/tests/karma.conf.js
```diff
@@ -35,24 +34,24 @@ module.exports = function(config) {
...
         loaders: [
           {
             test: /\.js$|\.jsx$/,
-            loader: 'babel',
+            loader: 'babel-loader',
             // Reason why we put this here instead of babelrc
             // https://github.com/gaearon/react-transform-hmr/issues/5#issuecomment-142313637
-            query: {
-              'presets': ['es2015', 'react', 'stage-0'],
-              'plugins': [
```
psimyn: Is this hmr thing still an issue?

psimyn:
images

PR #823, opened 2017-02-07
[DO NOT MERGE] proof of concept - Webpack dll

GGAlanSmithee:
This is a POC to follow up on my comment in https://github.com/reactGo/reactGo/issues/820#issuecomment-277822948

This is just to show one way of long-term caching, and maybe as a reference if you decide to do it this way. It was a rather quick implementation and probably not optimal

This does the following:

* adds npm scripts for build development and production (minified) dll build of some vendor libs
    - `npm run build:dll`
    - `npm run build:dev:dll`
* adds corresponding webpack config for producing and consuming the dlls
* loads vendor scripts dynamically using a manifest file (also depending on dev / prod)

(in a bit of a hurry, will answer any questions in the comments)


GGAlanSmithee:
Also, some caveats that I've noticed / have had problems with:

* it does not play well with `offline-plugin` (not too sure about this one, might be something I did wrong)
* shared libraries (by many dlls) are inclded many times
    - in this PR for example, I belive redux is in both `core` and `react` vendor files
    - not sure if this is fixable - am investigating this

choonkending:
@GGAlanSmithee Excellent! Thanks for this POC!

PR #826, opened 2017-02-10
Fix Model.associate not being called for sequelize

joonhocho:
Related: https://github.com/reactGo/reactGo/issues/825


joonhocho:
@choonkending this fixes https://github.com/reactGo/reactGo/issues/825.

choonkending:
@joonhocho Thanks for this! Will review when I get the chance to get it merged!

choonkending:
Tested and working. The reason why that bit of code wasn't working was because I was iterating through a list of keys (not an array of models)! Good find.

PR #837, opened 2017-03-06
remove react-addons-test-utils from package.json

GGAlanSmithee:
Seems like this package isn't used any more, guess it was superseeded by `enzyme` ?


GGAlanSmithee:
@choonkending You are correct! Sorry for the noise.

PR #847, opened 2017-03-18
Proposal: Use Mocha as a test runner with jsdom

choonkending:
# Why

**Simplify our test setup**

We have _a lot_ of moving parts involved when running our test suites. We use karma as a test runner, and as a consequence, we have heaps of other glue code from various dependencies:
- karma-jsdom-launcher
- karma-webpack, etc

# High Level Changes
- Bump babel-dependencies
- Use mocha as a test runner
     - Create a setup file which does the following:
         - babel-register
         - Initialises jsdom by loading `document` into the global scope. Reference [here](https://github.com/airbnb/enzyme/blob/master/docs/guides/jsdom.md)
- Remove karma dependencies
- Remove tests.webpack.js
- Remove karma.conf.js

Please let me know if you think there is something here that is lacking that karma supports!


choonkending:
I will be merging this tonight if there are no objections :)

app/tests/setup.js
```diff
@@ -0,0 +1,22 @@
+require('babel-register') ({
+  presets: ['es2015', 'react', 'stage-0']
```
psimyn: worth importing  these from `webpack/config/rules`?

app/tests/compilers/css.js
```diff
@@ -0,0 +1,4 @@
+function noop() { return null; }
+
+require.extensions['.css'] = noop;
```
psimyn: shoudl there also be an extension for .jsx?

current webpack config allows these to be imported without an extension, but they won't be picked up by tests. Also does this work for imports without extension?

app/tests/actions/topics-test.js
```diff
@@ -5,8 +5,9 @@ import md5 from 'spark-md5';
 import { polyfill } from 'es6-promise';
 import axios from 'axios';
 import expect from 'expect';
-import * as actions from 'actions/topics';
-import * as types from 'types';
+import sinon from 'sinon';
```
GGAlanSmithee: @choonkending is there any way to retain the relative paths? That would be :1st_place_medal:

ZeroCho: @GGAlanSmithee Paths like `actions/topics` and `types`? Currently [here](https://github.com/reactGo/reactGo/blob/master/webpack/resolve.js#L4)

GGAlanSmithee: @ZeroCho yeah, with the previous test setup, the webpack module resolve was used, allowing us to import components relative to webpacks module paths, but it looks like this was lost with this new setup.

ZeroCho: @GGAlanSmithee I'm not sure it is lost, but @choonkending agreed that we use '../../types' instead of 'types' because of possibility of duplication with npm package named 'types'.

GGAlanSmithee: @ZeroCho ah ok, my missunderstanding then :+1:

app/tests/compilers/css.js
choonkending: I don't believe we need a compiler for jsx. Mainly because we require babel-register in setup. My understanding is that it will babelify modules as we require them. The extensions we need to ignore are the ones we babel will not understand, e.g. css.

app/tests/actions/topics-test.js
choonkending: Good concerns @ZeroCho and @GGAlanSmithee . 

We aren't currently using webpack to retain the magic paths unfortunately. Although I find it really neat to do without nested `../../../`, I think in the future, I would favor moving the tests to within each component, similar to below:

```js
/component
   /__test__/
       - Component-test.js
   - Component.js
   - reducer.js
   - actions.js

```

My team and I have found it really neat and easy to locate components. Relative paths become less important when doing so. 

Hopefully, this doesn't become too much of a trade-off 🤞

app/tests/compilers/css.js
psimyn: i mean for resolving them, not for parsing.

if `some-test.js` uses `import thing from 'thing'`, then it won't find `thing.jsx` (but non-test code will).

choonkending: Wouldn't babel convert it into commonJS with our [preset](https://babeljs.io/docs/plugins/preset-es2015/) and from there NodeJS will just require it?

I put a console log in one of our jsx files that are tested and it was logged.

app/tests/setup.js
choonkending: Oh you @psimyn

choonkending: Hmm I thunk about it a bit further. 

Currently `webpack/rules/javascript`  is a function that accepts `production` or `browser` flags, which also returns `{}` that contains the rules required by webpack.

```js

var javascript = require('../../webpack/rules/javascript');
var babelOptions = javascript({ browser: true, production: true }).options;

require('babel-register') (babelOptions);
```

I initially intended to just use `webpack/rules` as an entry point (or interface) and did not see the need for javascript to be exposed outside. As a result, the javascript rules we are using is very coupled to webpack/rules. 

I'm not sure if accessing this `{}.options` is a good idea, because we are tightly coupling our  mocha setup to a webpack-specific rule, i.e. if webpack rules decide to accept something other than `option`, this setup would break. 

What we want is to actually use the common babel config that will be used in `webpack/rules/javascript` (which will transform the babel config into what webpack loves) as well as `setup.js`.

A logical place for a babel config file which exports presets and plugins should be `config/babel.js`.  However, I think this will only make the configuration wayyy more confusing. :(

What do you think? #overengineering?

app/tests/actions/topics-test.js
GGAlanSmithee: @choonkending yeah I guess that is the logical evolution of componentalizing the project structure. This seems to be the way the winds are blowing in react land with colocation of view, data fetching, styles etc, so why not do the same with tests!

app/tests/setup.js
choonkending: I think it's okay to keep as-is currently. If we find that we need to do it later, we will tackle that in a separate PR.

choonkending:
Mergingggggg in 10 minutesss

PR #853, opened 2017-04-08
Create boundary around axios

choonkending:
# Why

As part of my investigation into solving https://github.com/reactGo/reactGo/issues/835

Creating a boundary around our HTTP client gives us more control over our third party dependencies. 

By defining our own interface (even if it is similar to axios), we have less places in our code that depend on axios directly. It's better to depend on something you control rather than something you don't control, lest it end up controlling you.

# High Level Changes

- Created a RestApiClient that exposes a `request` method
- Tested it by actually spinning up an express server and inspecting the request
   - Involved modifying jsdom setup

## Thoughts
1. I think testing this client involves testing the "output" or what you "expect" from calling this interface, which in my mind are two things:
- A request to the server
- A Promise returned to us

I did not test the "promise" returned, mainly because I am unsure whether it is worth testing in the first place, but also because I am thinking of making the rest api client return a "Task", which in FP land is a handy way of handling asynchronous code.

2. I opted to keep the test simple. But can add more tests (get, put, delete) if anyone deems it necessary.


slavab89:
I actually did something very similar to this on my end.
I've created a wrapper like you. But in order to add the ability to handle authenticated requests on the server side, you need the cookie + other headers (possibly).
For example, if you access to the DB is checking `req.isAuthenticated()` then it will always fail on the server-side rendering part because you dont have the cookie present when the action uses `Axios` to send the actual API request.
To fix this i've added the following pieces of code:

```
// restApiClient.js
initializeAPIClients(ssrHeaders) {
  if (typeof window === 'undefined' && ssrHeaders) {
    Object.assign(this.client.defaults.headers.common, ssrHeaders);
  }
}
```
```
// client.jsx
const store = configureStore(initialState, browserHistory);
const history = ...
const routes = ...
restApiClient.initializeAPIClients();
...
```
```
// middleware.js
...
const ssrHeaders = {
  'User-Agent': req.headers['user-agent']
};

if (req.cookies.sessionId) { // Need to use cookie-parser module
  ssrHeaders.Cookie = `sessionId=${req.cookies.sessionId}`; // sessionId is the name of the passport cookie
}
restApiClient.initializeAPIClients(ssrHeaders);
const routes = createRoutes(store)
```

The structure might differ a little and might be places at a different place but basically, when doing ssr requests, it should pass the user-agent (might be useful) and the passport cookie along the request

choonkending:
@slavab89 Thanks for that. Agreed! This was my sort-of end goal. First create a boundary around our rest API client, and then conditionally add headers as you did. 

Going off tangent a bit, might be worth looking at using [JSON web tokens](https://jwt.io/) for authentication rather than passport. Have you used this before?

GGAlanSmithee:
@choonkending adding jwt is on my to-do list too, and I've briefly looked at it. I thought you would (or atleast could) use it in conjunction with passport? For example https://github.com/kriasoft/react-starter-kit/blob/master/src/server.js

choonkending:
Made progress on this:

- Used service module in our action creators instead of accessing axios directly. Refactored the tests to have a stub service function. Still not happy with the fact that we have to stub or mock our tests, but I see this as an improvement. Read more about why stubbing/mocking isn't ideal [here](http://rea.tech/to-kill-a-mockingtest/). Do not have a clear solution in my head to make our tests + code more modular + testable yet.

- Remove references of `axios` in our codebase (except for actions/user, which I plan to tackle in our next PR).

Feels like this is a good improvement peeps! Will merge shortly and have a follow up PR for actions/user!

PR #855, opened 2017-04-08
Yarn && React@15.5.0 && React-helmet@5

ZeroCho:
@choonkending 
#821 

I updated react-router to v3 and react-helmet to v5
I failed to update react-router to v4. Hard to achieve server side rendering


ZeroCho:
test needed for postgreSQL, as I only tested it working with Mongo

choonkending:
Thanks for this!

I've heard good things about yarn. And good upgrade to react 15.5! The changes look good.

I will have to look into the lock file and see if it makes sense (I haven't used Yarn before).

Were the changes for eslint a result of fixing our eslint problems? I like not `()` in our arrow functions lol, but consistency is good!

Were there any gotchas when upgrading to react 15?

choonkending:
I will also write a Wiki around upgrading to Yarn and why we decided to do so. If you have found any good points/advantages would be good to list them so we can add that to our documentation. It will be good for anyone who follows the progress of reactGo and have a bit of insight into our tech decisions!

Thanks @ZeroCho for always being around to help with the hard issues!

ZeroCho:
@choonkending There is no breaking change in react 15.5. It is just for preparing react 16.

ZeroCho:
@choonkending There is no breaking change in react 15.5. [Release note](https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html)
It is just for preparing for react 16.

Yarn is compatible with npm. Just install yarn and type `yarn` in cli, everything will be automatically processed.

slavab89:
I have yet to get experienced with yarn but it does look promising.
A few notes that i've found on it:
* Installed packages are cached offline under `~/.yarn-cache`
* Creation of a `lock` file to have deterministic installs - will always install the same version (similar to Shrinkwrap)
* It seems that yarn is quite faster than npm during installation of packages - might be related to the fact that they use a different package site as well: `https://registry.yarnpkg.com`.

Didnt really find major drawbacks, and starting to look at the 700+ issues there is in the Yarn repository seems too much for me :P

ZeroCho:
yarn is fully compatible with npm, so I just wanted to give users another
option.

2017. 4. 10. 오후 5:48에 "slavab89" <notifications@github.com>님이 작성:

choonkending:
LGTM

choonkending:
@ZeroCho Thanks for this. Excellent work.

ZeroCho:
@choonkending Heroku also support Yarn as of march 2017!

slavab89:
Regarding the SSR and V4 of react-router.
Might some new development?
https://github.com/ReactTraining/react-router/issues/4892#issuecomment-298444543
https://github.com/ReactTraining/react-router/issues/4977
Really have no idea whats up with [react-router-redux](https://github.com/ReactTraining/react-router/tree/master/packages/react-router-redux) package though. There is an alpha.6 now (Where do i see it?)

PR #859, opened 2017-04-16
broken link (aws.md is lowercase)

avloss:


choonkending:
@avloss Thanks for that!

PR #867, opened 2017-04-30
Long term caching - Episode 0

choonkending:
# Why

Part of https://github.com/reactGo/reactGo/issues/820#issuecomment-298142272

As an initial step, before we work towards code splitting.

# High Level Changes
- Separate build task for webpack production to individual webpack tasks
- Use webpack-manifest-plugin to generate manifest
    - Use generated `manifest.json` when creating script and stylesheet tags
- Use chunkhashes for filenames and stylesheets in production
- Create `static-assets` module that generates assets for dev and production mode
- Add `__PRODUCTION__` compile time constant to webpack for minification in conditional requires


choonkending:
Merging in a few hours, get your comments in!

Issue #848, opened 2017-03-26
Webpack loader issue - DeprecationWarning: loaderUtils.parseQuery()


slavab89:
When running the project i see the following warning:
```
(node:19056) DeprecationWarning: loaderUtils.parseQuery() received a non-string value which can be problematic, see https://github.com/webpack/loader-utils/issues/56
parseQuery() will be replaced with getOptions() in the next major version of loader-utils.
```
I've tried to add `process.traceDeprecation = true` to various places but for some reason it doesnt show which loader causes this issue.

Anyone knows how to fix this?

ZeroCho:
One of loaders must be using `parseQuery` instead of `getOptions` but I searched all loaders(postcss, babel, style, json, css, url, file) I could not find `parseQuery`. That's very weird.

slavab89:
If i search for `./parseQuery` in all node_modules the almost all loaders have a require to this. But i'm not sure if this is it - cause its in all loaders

slavab89:
So, i've looked at this a bit further.
Seems to be happening on [babel-loader](https://github.com/babel/babel-loader/issues/392) plugin
It is fixed on version 7 that's on beta there.

So nothing much that we can do here other than wait =]

GGAlanSmithee:
That's been bugging me too, thanks for diggin @slavab89

slavab89:
So... getting back to this. Just after 2 days since i've close this issue. They've released version 7 of babel-loader.

I've updated to the latest version of all loaders and dont see this happen anymore

zoruc:
@slavab89 I did the same thing and it fixed the issue thanks

PR #875, opened 2017-05-12
update babel-loader

### LINKED ISSUE #874: Webpack deprecation error
## Deprecation
When running npm run dev

**DeprecationWarning: loaderUtils.parseQuery() received a non-string value which can be problematic, see https://github.com/webpack/loader-utils/issues/56
parseQuery() will be replaced with getOptions() in the next major version of loader-utils.**

## Question

How do you get rid of this warning? I've updated all the packages and loaders to points even further along than what you had in and I still get them is this something that is not configured correctly in the webpack configs

GGAlanSmithee:
Fixes #874 and #848 by updating `babel-loader`.

On another note, there are more deps to be bumped, with new major, minor and patch versions. I have done so in my own projects (except for `react-router`), and haven't noticed any regressions. Running `ncu` ([npm-check-updates](https://www.npmjs.com/package/npm-check-updates)) yields

```javascript
Using /home/ubuntu/workspace/package.json

 autoprefixer                                     ^6.7.7  →   ^7.0.1 
 babel-plugin-transform-react-remove-prop-types   ^0.3.3  →   ^0.4.5 
 cross-env                                        ^4.0.0  →   ^5.0.0 
 postcss-loader                                   ^1.3.3  →   ^2.0.5 
 react-router                                     ^3.0.3  →   ^4.1.1 
 style-loader                                    ^0.16.1  →  ^0.17.0 
 eslint-plugin-jsx-a11y                           ^4.0.0  →   ^5.0.1 
 eslint-plugin-react                             ^6.10.3  →   ^7.0.0 
 jsdom                                           ^9.12.0  →  ^10.1.0 
 redux-mock-store                                  1.2.2  →    1.2.3 

The following dependencies are satisfied by their declared version range, but the installed versions are behind. You can install the latest versions without modifying your package file by using npm update. If you want to update the dependencies in your package file anyway, run ncu -a.

 prop-types          ^15.5.4  →  ^15.5.9 
 react-router-redux   ^4.0.0  →   ^4.0.8 
 mocha                ^3.2.0  →   ^3.3.0 

Run ncu with -u to upgrade package.json
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "axios": "^0.16.0",
     "babel-cli": "^6.24.1",
     "babel-core": "^6.24.1",
-    "babel-loader": "^6.4.1",
+    "babel-loader": "^7.0.0",
     "babel-plugin-react-transform": "^2.0.2",
     "babel-plugin-transform-react-constant-elements": "^6.23.0",
     "babel-plugin-transform-react-inline-elements": "^6.22.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -459,14 +459,13 @@ babel-helpers@^6.24.1:
     babel-runtime "^6.22.0"
     babel-template "^6.24.1"
 
-babel-loader@^6.4.1:
-  version "6.4.1"
-  resolved "https://registry.yarnpkg.com/babel-loader/-/babel-loader-6.4.1.tgz#0b34112d5b0748a8dcdbf51acf6f9bd42d50b8ca"
+babel-loader@^7.0.0:
+  version "7.0.0"
+  resolved "https://registry.yarnpkg.com/babel-loader/-/babel-loader-7.0.0.tgz#2e43a66bee1fff4470533d0402c8a4532fafbaf7"
   dependencies:
     find-cache-dir "^0.1.1"
-    loader-utils "^0.2.16"
+    loader-utils "^1.0.2"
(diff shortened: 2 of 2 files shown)


psimyn:
it's ok to keep `babel-core` etc at 6.x?

choonkending:
Thanks for this PR! Also worth updating the Babel deps along the way as @psimyn suggested

GGAlanSmithee:
@choonkending @psimyn the other babel related packages have not had their versions bumped (see https://www.npmjs.com/search?q=babel).

Regarding yarn, it would be great if you want to do it @choonkending. It's probably trivial but I havn't touched yarn yet :)

psimyn:
@GGAlanSmithee are you able to update the yarn lockfile as well? Will require installing yarn, described on [their installation page](https://yarnpkg.com/lang/en/docs/install/)

then you can use

```
yarn upgrade babel-loader
```

`yarn upgrade` will update the lockfile _and_ the package.json - means they both stay correct versions. Other than that it's very similar to using npm, but generally faster.

Yarn command cheat sheet - https://shift.infinite.red/npm-vs-yarn-cheat-sheet-8755b092e5cc#2837

Thanks!

GGAlanSmithee:
@psimyn updated, thanks for the info!

PR #878, opened 2017-05-17
linkAssets not returning causing favicon to not load

andrewmoon1:
Added return to linkAssets function in order for favicons to load properly.


choonkending:
Good find! @andrewmoon1 ! Thanks!

PR #891, opened 2017-05-30
Wrap ${componentHTML} to kill checksum warning

jrodl3r:
Wrapping ${componentHTML} with an empty &lt;div&gt; gets rid of this console warning:

*Warning: React attempted to reuse markup in a container but the checksum was invalid*

### DIFF
--- a/server/render/pageRenderer.jsx
+++ b/server/render/pageRenderer.jsx
@@ -23,7 +23,7 @@ const buildPage = ({ componentHTML, initialState, headAssets }) => {
     ${staticAssets.createTrackingScript()}
   </head>
   <body>
-    <div id="app">${componentHTML}</div>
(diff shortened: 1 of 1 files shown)


GGAlanSmithee:
This seems wrong to me, as it would remove any re-use of ssr code? Is it in both prod and dev you get these messages?

jrodl3r:
Correct. That's a bad solution, and this issue should probably be renamed to "Fix Checksum Console Warning"

[Reading this](https://stackoverflow.com/questions/33521047/warning-react-attempted-to-reuse-markup-in-a-container-but-the-checksum-was-inv) has be wondering if it might be this:

[*"Your issue appears to be with CSS - if you are using autoprefixer and inline-styles that would explain your difference here."*](https://stackoverflow.com/a/40847348)

Issue #888, opened 2017-05-26
Populate User State After Authentication


jrodl3r:
What is the suggested method for loading user data (ex: `email`, `fname`) after a successful login?

Should it be fetched after POST, or should the controller pass an object back on successful `passport.authenticate(...)`?

Thx 👍

slavab89:
I would say that after a successful login `passport.authenticate` you should redirect to your home page.
When that happens, it would trigger an SSR in [middleware.js](https://github.com/reactGo/reactGo/blob/master/server/render/middleware.js#L16) where the creation of a new store would happen.
As you can see there, the user details are initialized so if `req.isAuthenticated()` returns true, you can grab the user details that you've saved on the session by using `req.user` and save them / or some of them in the store

jrodl3r:
Thanks for the tip @slavab89, but one thing I don't understand is why anyone would ever redirect a user to the home page after a successful login? I can't recall any site that has ever done that before, because they all send to your /account page as a logged-in user typically.

jrodl3r:
@choonkending @psimyn what do you suggest for fetching and persisting user data securely on/after login? Is this possible right now?

slavab89:
I'm doing the following: (For some reason its not really documented in passport docs)
```js
    app.get('/auth/callback',
      passport.authenticate('oidc', { failWithError: true }),
      (req, res) => {
        res.redirect('/');
      },
      (err, req, res, next) => { // Special specific error callback for that route
        res.redirect('/?loginError=' + encodeURIComponent(err));
      }
    );
```

But it should work with the current code that's in the branch.

And in middleware.js
```js
  const store = configureStore({
    user: {
      authenticated,
      userData: req.user,
      ......
    },
    ......
  }, history);
```

It should trigger a SSR request because you force the user to do a redirect to your home page. You could add a print in `middleware.js` to see if it actually reaches there...

ZeroCho:
I populate them in `deserialize` function like below
export function deserializeUser (id, done) => {
  User.findById(id).populate('friendRequest friends').exec((err, user) => {
    done(err, user);
  });
};

jrodl3r:
You guys are awesome, really appreciate the help on this! Feels so close to working... However, still coming up empty handed on both options...

[...]

PR #892, opened 2017-05-30
setup user profile state + simple data population

jrodl3r:
Addresses #888


trda:
Perfect. Thanks a lot for the solution.

PR #898, opened 2017-06-13
adds package-lock.json

GGAlanSmithee:
This adds the `package-lock.json` that is generated when installing packages with node >= 8 and npm >= 5. Works pretty much like yarn.lock and should be comitted by advice of npm log.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,6961 @@
+{
+  "name": "react-webpack-node",
+  "version": "3.4.1",
+  "lockfileVersion": 1,
+  "dependencies": {
+    "@types/geojson": {
+      "version": "1.0.2",
+      "resolved": "https://registry.npmjs.org/@types/geojson/-/geojson-1.0.2.tgz",
+      "integrity": "sha1-sC0QqwKOKSisWSoFGqpJgaGUHQM="
+    },
+    "abab": {
+      "version": "1.0.3",
+      "resolved": "https://registry.npmjs.org/abab/-/abab-1.0.3.tgz",
+      "integrity": "sha1-uB3l9ydOxOdW15fNg08wNkJyTl0=",
+      "dev": true
+    },
+    "abbrev": {
+      "version": "1.1.0",
+      "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.0.tgz",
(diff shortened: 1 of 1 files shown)


choonkending:
Thanks @GGAlanSmithee !

GGAlanSmithee:
I am having some troubles with `package-lock.json`, where updated dependencies does not get installed. Apperantly, it's a known "issue", https://github.com/npm/npm/issues/16866, but it seems like it's by design. 

My opinion is that the behaviour is rather non-intuetive. If `package.json` is updated, it should reflect in the `package-lock.json` file, but I understand why it doesn't, since it kind of defeat the purpose of it.

One way to mimic the old (`npm v ^4`) behaviour is to add a script; `rm -f package-lock.json && npm install`, which is kind of a hack, but gets the job done.

What do you guys think? Is this a non-issue? Should it be handled by `npm`? Should we supply a script to forcefully update the dependencies?

ZeroCho:
@GGAlanSmithee @choonkending Another issue here. Currently, we have both package-lock.json and yarn.lock. This is not recommended for some CI tools such as codeship and circle CI.
They get confused what to use to install packages, yarn or npm.
I think we need to remove one of them.

choonkending:
I think it makes sense to just use yarn.lock as a single source of truth. That might help eliminate the problem with the tooling.

GGAlanSmithee:
Yeah, probably. The issue is that npm v5 will add `package-lock.json` automatically everytime you do `npm install`. To avoid it, you have to do `npm set package-lock false`, but I don't think you can add that locally, as in `package.json`. If it is possible to do so, we probably should, but otherwise, all we can really do is to remove it and add it to `.gitignore`.

choonkending:
Do you think it makes sense if we just mention that we should `yarn install` instead of `npm install`? I'm a bit cautious of adding `.gitignore` for newbies wondering why their package-lock isn't committed in git.

ZeroCho:
@choonkending Yes. `yarn install` **only**
Also good to mention that if newbies don't wanna use `yarn`, they should remove `yarn.lock` and then `npm install`

GGAlanSmithee:
@ZeroCho @choonkending agreed! `yarn install` is IMO doing the most correct (as in intuetive) thing. Should we remove `package-lock.json`, as it will (probably?) be outdated as we upgrade deps?

PR #899, opened 2017-06-13
Sequelize upgrade

GGAlanSmithee:
Upgrades sequelize. There is no longer a `classMethods` or `instanceMethods` property, but rather you append the methods to the model or its prototype respectively.

This PR is based on #898 but not logically depending on it.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,6961 @@
+{
+  "name": "react-webpack-node",
+  "version": "3.4.1",
+  "lockfileVersion": 1,
+  "dependencies": {
+    "@types/geojson": {
+      "version": "1.0.2",
+      "resolved": "https://registry.npmjs.org/@types/geojson/-/geojson-1.0.2.tgz",
+      "integrity": "sha1-sC0QqwKOKSisWSoFGqpJgaGUHQM="
+    },
+    "abab": {
+      "version": "1.0.3",
+      "resolved": "https://registry.npmjs.org/abab/-/abab-1.0.3.tgz",
+      "integrity": "sha1-uB3l9ydOxOdW15fNg08wNkJyTl0=",
+      "dev": true
+    },
+    "abbrev": {
+      "version": "1.1.0",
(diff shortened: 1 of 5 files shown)


ZeroCho:
@GGAlanSmithee I guess @choonkending is busy these days

GGAlanSmithee:
@ZeroCho yeah, such is the life of the busy programmer :) Maybe someone else has the power to merge PRs? It's quite small and shouldn't be to impactful.

ZeroCho:
@GGAlanSmithee Actually, I have, but I'm familiar with PostgreSQL. I think at least one people should review each PR. Hopefully, @choonkending will be back soon.

choonkending:
Have been feeling a bit ill the past week. Plus presentations at meetups + blogposts [here](http://rea.tech/introduction-to-option-and-either-types-in-js-part-1/). :D

GGAlanSmithee:
@ZeroCho @choonkending nice! Hope I didn't come off as pushy. You will merge when / if you want too :)

Here is a changelog for Sequelize v4 http://docs.sequelizejs.com/manual/tutorial/upgrade-to-v4.html. There shouldn't be any breaking changes for this probject as far as I'm aware, but there might be some problems with `DATETIME` fields, as they are now represented as a new SQL type. It is possible to migrate them though, see https://github.com/sequelize/sequelize/pull/7201#issuecomment-278899803, maybe I should add that as a migration?

choonkending:
No worries!

I think if the DATETIME fields are not specified in this project (unless they affect the existing timestamps?), we shouldn't need to worry about them?

GGAlanSmithee:
@choonkending You are right! I am not sure what I was thinking;

1. There are no `DATETIME` fields in any model, only `DATE` fields
2. The breaking change was for `DATETIME2`, not `DATETIME`

:laughing:

GGAlanSmithee:
Updated `yarn.lock` and `package-lock.json`.

Issue #902, opened 2017-06-27
Webpack 3


GGAlanSmithee:
### Upgrade to Webpack 3

Webpack 3 [has been released](https://medium.com/webpack/webpack-3-official-release-15fd2dd8f07b). Most interesting it introduces rollup style scope hoisting via the `webpack.optimize.ModuleConcatenationPlugin`. 

Scope hoisting means that the scope of each module will be hoisted to the top of the bundle, potentially greatly reducing size and parse time of the code.

Even if it's a major release, it shouldn't introduce any breaking changes, so I will make a PR when I get time, unless someone beats me too it.

ZeroCho:
😢 `webpack.optimize.ModuleConcatenationPlugin` doesn't decrease the file size as much as I expected, but still I think it's good for ReactGo to apply that option.

GGAlanSmithee:
@ZeroCho yeah true. From what I can see the size goes from **337.32KB** to **332.46KB** so that is certainly a modest amount. Reportedly, there is also the benefit of faster browser evaluation time. I haven't had time to test this though, and I think the application is really too small of a sample to get any conclusive eveidence. 

Having said that, I don't think there's any appearant drawbacks to upgrading!

GGAlanSmithee:
#903

PR #903, opened 2017-06-27
Upgrade to webpack 3

GGAlanSmithee:
See #902

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -70,14 +70,14 @@
       }
     },
     "ajv": {
-      "version": "5.1.5",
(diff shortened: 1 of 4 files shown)


choonkending:
@GGAlanSmithee Holy, that was really easier than I expected!

choonkending:
Did you have any problems around `extract-text-webpack-plugin` upgrade? Got a warning but still seems to be working

slavab89:
Did you meann the PEER DEPENDENCY?
It seems that they are trying to solve it over there https://github.com/webpack-contrib/extract-text-webpack-plugin/issues/561

PR #916, opened 2017-08-07
Fixed build script for windows.

edmiester777:
Windows would throw "NODE_ENV is not recognized as an internal or external command" when running build script. This was due to a missing cross-env command.


choonkending:
LGTm!

choonkending:
Thanks @edmiester777 !!

Issue #908, opened 2017-07-10
Code Splitting?


ramakay:
## Enhancement

** Initial app.js cost is high.

SSR is great but `app.js` is a big payload and [startrender](https://sites.google.com/a/webpagetest.org/docs/using-webpagetest/quick-start-quide#TOC-Start-Render:) is adversely impacted. 

Options:

1. React Router to dynamically load chunks , however SSR efforts were incomplete

https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/docs/guides/code-splitting.md#code-splitting--server-rendering

Possible solution:

https://blog.emilecantin.com/web/react/javascript/2017/05/16/ssr-react-router-4-webpack-code-split.html 

2. Webpack dynamic imports

Does not seem to be straight forward. 

https://medium.com/webpack/how-to-use-webpacks-new-magic-comment-feature-with-react-universal-component-ssr-a38fd3e296a

Sadly, the synchronous nature of SSR - This whole things seems challenging, thoughts?

Apologies, if this was discussed earlier - please point me to the description

slavab89:
There are a few PR's that try to achieve this
https://github.com/reactGo/reactGo/pull/841 (Does code split in RR3)
https://github.com/reactGo/reactGo/pull/880
But all seem to reach a dead end for now 😢 Its the combination or RR4, SSR, code split, hot reloading. Seems like its possible to achieve some of them but not all together.

If you have some additional insights or trials & experience that you've tried it would be great to know.

ZeroCho:
Actually I succeeded in code splitting + server side rendering with RR4, but it has some breaking changes.
I think reactGo should adopt that in its next version.

slavab89:
😱 Looking forward for this so much!! *\*Heavy Breathing\** 😆

ramakay:
@slavab89 seems to be the way to go - split async and sync bundles and use webpack to replace at runtime. I am doing that with reactgo. 

https://blog.emilecantin.com/web/react/javascript/2017/05/16/ssr-react-router-4-webpack-code-split.html

@ZeroCho  - please share where you reached and if your approach was similar by using the `NormalModuleReplacementPlugin` 

Thanks guys, lets work towards this 👍

ramakay:
Np, continue to play with the NormalModuleReplacementPlugin, it lead me to a necessary upgrade to Router 4 - what a drain :(

slavab89:
Could one of you please share what you've achieved so far if possible?

[...]

PR #841, opened 2017-03-07
Code split

GGAlanSmithee:
Enables code splitting by using dynamic entry points in `routes.jsx`

high level features:

* enables code splitting on client
    - hashed chunks in prod mode (for caching and cache-busting)
    - named (id) in dev mode
* disables code splitting on server
* uses a webpack manifest file to make sure vendor chunks does not change unless code changes
* adds a vendor chunk as an example (the actual content of this can easily be changed though)

Please ask any questions regarding the implementation you might have.

PR #919, opened 2017-08-09
react-router v4 && react-hot-loader@3 && Code Splitting

ZeroCho:
@choonkending @slavab89


ZeroCho:
@GGAlanSmithee #841

ZeroCho:
#908

Cleanshooter:
I'll do a fresh pull to night and check this out! <-- excited 😄

GGAlanSmithee:
@ZeroCho Thanks for the ping! Very exciting, I will take a look at it as soon as I get some free time. Without having looked at it - does this PR do server-side rendering + code splitting with `react-router@4`? If so, that's very nice, I have struggled with it a lot. Been a bit intrigued by https://github.com/faceyspacey/redux-first-router, wanting to test it out, have you come across it?

ZeroCho:
@GGAlanSmithee Yes. SSR + Code splitting with react-router@4
I didn't have time to comment on my code, and didn't test it, so errors may happen

choonkending:
@ZeroCho This is an excellent! 

I like the fact that we are no longer relying on react-router-redux any longer. Decoupling our navigation from redux was one of the aims I had in mind.

I would like to spend some time this weekend playing around further with this branch in order to have an understanding of what's changed. Perhaps some modules can be split into more, mainly because they seemed to be doing too many things. But I would have to try it out in order to give a better comment!

ZeroCho:
@choonkending Decoupling `react-router-redux` is quite easy. If you want me to decouple it. I will.

However, I like `react-router-redux` because it dispatches `LOCATION_CHANGE` action everytime url changes. By listening to the action, I can simulate `onUpdate` of react-router@3, which is in `app/client.js`

package.json
```diff
@@ -20,90 +20,91 @@
...
     "spark-md5": "^3.0.0",
-    "style-loader": "^0.16.1",
-    "url-loader": "^0.5.8",
-    "webpack": "^2.3.3",
-    "webpack-dev-middleware": "^1.10.1",
-    "webpack-hot-middleware": "^2.18.0",
-    "webpack-manifest-plugin": "^1.1.0"
+    "style-loader": "^0.18.2",
+    "url-loader": "^0.5.9",
+    "webpack": "^3.5.2",
```
choonkending: Wow! @ZeroCho - You've also updated webpack.  Bloody champion

slavab89:
So i've started to play around with it on my side and i'm getting
```
Warning: setState(...): Can only update a mounting component. This usually means you called setState() outside componentWillMount() on the server. This is a no-op. Please check the code for the AsyncComponent component.
```
I presume that this happens because of the promise there.. `componentWillMount` exists and when the promise returns `setState` is called.

How would it be possible to disable the async component loading on the server side? Doesnt seem to be necessary anyway, no? Maybe with something like https://github.com/thejameskyle/react-loadable ?

ZeroCho:
@slavab89 Changing from `componentWillMount` to `componentDidMount` will resolve that problem.

GGAlanSmithee:
Been testing this a bit and it seems to be working fine, with the small aside mentioned by @slavab89. I definitely agree that code splitting on the server should be avoided if possible. I have no idea how to do this with `AsyncComponent` though. Using webpacks chunk plugin, i managed it with a hack;

`new webpack.optimize.MinChunkSizePlugin({minChunkSize: browser ? 10000 : Infinity})`

@ZeroCho 

> And it is for Search Engine Optimization.

Care to elaborate on this? Not sure what you are reffering to exactly.

ZeroCho:
@GGAlanSmithee Ah, I now got what @slavab89 means. I did some misunderstanding. I think I can change some code of `app/routes.js` to avoid code splitting on the server.

```
component: isServer ? Home : asyncComponent(...)
```

slavab89:
So i've just discovered something that might bring us the solution..?
First of all, the article: https://medium.com/faceyspacey/code-cracked-for-code-splitting-ssr-in-reactlandia-react-loadable-webpack-flush-chunks-and-1a6b0112a8b8
It is actually pretty interesting.
And the module itself (Pretty much react-loadable successor) https://github.com/faceyspacey/react-universal-component
And they've create a demo as an example: https://github.com/faceyspacey/universal-demo

Did anyone of you heard about this?

app/containers/Dashboard.jsx
```diff
@@ -1,10 +1,32 @@
...
+import { connect } from 'react-redux';
 /*
  * Note: This is kept as a container-level component,
  *  i.e. We should keep this as the container that does the data-fetching
  *  and dispatching of actions if you decide to have any sub-components.
  */
-const Dashboard = () => <div>Welcome to the Dasboard. Stay tuned...</div>;
+class Dashboard extends Component {
+  componentWillMount() {
+    const { user: { authenticated }, history } = this.props;
```
GGAlanSmithee: Is it possible to couple this with route config as in rr3? Or is this the new way of protecting routes? If so, maybe this should be in it's own component (or HoC)? Usage would be like

```javascript
<Protected redirect={'/login'}>
    <MyComponent/>
</Protected>
```

app/routes.jsx
```diff
@@ -1,40 +1,46 @@
...
-      <IndexRoute component={Vote} fetchData={fetchVoteData} />
-      <Route path="login" component={LoginOrRegister} onEnter={redirectAuth} />
-      <Route path="dashboard" component={Dashboard} onEnter={requireAuth} />
-      <Route path="about" component={About} />
-    </Route>
-  );
-};
+}
+
+export default [{
```
GGAlanSmithee: a bit sad to see the jsx route trees go, was visually intuetive too me, but I guess that's the way the cookie crumbles.

GGAlanSmithee:
Took a brief look at the code. Not too much to add unfortunatley, it looks fine too me.

The only real issue discovered so far is the server-side code splitting if I understand it correctly? It might be inconveniant, but it's not really a bug. This PR seems like the way to go. `rr4` and `react-hot-loader 3` is :+1:. Maybe if it's merged, we can iterate on it? Is there anything besides time blocking it? :smile:

choonkending:
Hey peeps, just an update,  Ive been working on a branch inspired by @ZeroCho 's work here. It's still in WIP (lots to cover) https://github.com/reactGo/reactGo/compare/master...choonkending:react-router-4-clean.

I thought there might be a better way around our server side and client side data fetching, so I was experimenting around with that. 

Just to let you peeps know that I'm not ignoring this awesome PR!

slavab89:
So i've recently tried to use https://github.com/faceyspacey/react-universal-component to achieve SSR + Code Split. I can say that its really easy to do it with it. Just follow their demo.
The thing is that it still doesnt play super nice with RR4 (Which seems to be the standard 🤔 ) because of store and history syncing, but they also have https://github.com/faceyspacey/redux-first-router which works with it.
Basically they have a whole stack of projects intended for all of this.

What do you guys think? Or do we have some progress on this?

choonkending:
@slavab89 I'm still on my spike branch based off this one https://github.com/reactGo/reactGo/compare/master...choonkending:react-router-4-clean which I think has an acceptable solution to rr4, but without code splitting yet. I was thinking that could be tackled once we have a solution for this first! :D

app/client.jsx
```diff
@@ -1,49 +1,43 @@
...
 
-const store = configureStore(initialState, browserHistory);
-const history = syncHistoryWithStore(browserHistory, store);
-const routes = createRoutes(store);
+const renderDOM = (r) => {
+  render(
+    <AppContainer>
+      <Provider store={store}>
+        <ConnectedRouter history={history}>
+          <PendingNavDataLoader routes={routes} store={store}>
```
GGAlanSmithee: should this be 

```javascript
<PendingNavDataLoader routes={r} store={store}>
```

?

webpack/rules/javascript.js
```diff
@@ -1,18 +1,12 @@
...
 
 module.exports = ({ production = false, browser = false } = {}) => {
-  const enableHotModuleReplacement = !production && browser;
-  const createPresets = enableHotModuleReplacement => {
-    const presets = ['es2015', 'react', 'stage-0'];
-    return enableHotModuleReplacement ? ['react-hmre', ...presets]: presets;
-  };
-  const presets = createPresets(enableHotModuleReplacement);
-
+  const presets = ['es2015', 'react', 'stage-0'];
```
GGAlanSmithee: should this be 

```javascript
const presets = [
    ['es2015', {modules: false}],
    'react',
    'stage-0'
]
```

to allow for `webpack` dead code elimination?

app/client.jsx
ZeroCho: Good catch. You're right

webpack/rules/javascript.js
ZeroCho: @GGAlanSmithee Does that work in client environment too?

GGAlanSmithee: @ZeroCho I don't see why it shouldn't. The webpack bundling (and therefor code elimiation) happends at "compile" time, so it shouldn't matter in what environment the final code is used. Is there any specific reason that makes you think it wouldn't work in browsers?

ZeroCho: @GGAlanSmithee You're right. I checked out that it works in client too

app/containers/Dashboard.jsx
GGAlanSmithee: Here is an example of a HOC doing this:

```javascript
import {Component} from 'react'
import PropTypes from 'prop-types'
import {connect} from 'react-redux'

const asProtected = (WrappedComponent, selectData) => {
    class Protected extends Component {
        static propTypes = {
            user: PropTypes.object,
            history: PropTypes.object
        }
        
        componentWillMount() {
            const {user: {authenticated}, history} = this.props
            
            if (!authenticated) {
                history.replace('/login')
            }
        }

        render() {
            return <WrappedComponent {...this.props}/>
        }
    }
    
    const mapStateToProps = ({user}) => ({
        user
    })
    
    return connect(
        mapStateToProps
    )(Protected)
}

export default asProtected
```

Usage:
```javascript
import asProtected from 'hoc/asProtected'

const MyComp = () => (
    <div>
        Only for logged in users!
    </div>
)

export default asProtected(MyComp)
```

GGAlanSmithee: The same could probably be done for redirections.

PR #920, opened 2017-08-10
Reduce time to DOMContentLoaded by loading the javascript asynchronously

alubbe:
Because of the serverside rendering, there is no reason to wait for the entire JS to be downloaded, parsed and executed before rendering the site on the client. Adding `async` is the current best practice to achieve that.

Using `async` in dev has no effect on the development experience as far as I can tell. The CSS is already flashing on each reload and the load times are the same. So strictly speaking, it would be enough to add `async` to production only, but I prefer to have the dev env be as similar to the prod env, so that bugs and race conditions can be discovered there.

### DIFF
--- a/server/render/static-assets/dev.js
+++ b/server/render/static-assets/dev.js
@@ -1,6 +1,6 @@
 import { GOOGLE_ANALYTICS_ID } from '../../../config/env';
 
-const createAppScript = () => '<script type="text/javascript" charset="utf-8" src="/assets/app.js"></script>';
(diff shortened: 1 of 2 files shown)


alubbe:
Also, feel free to squash these commits - I made these changes on the github edit interface where I can't fit more than one file into a commit :)

ZeroCho:
LGTM! I've already put that `async` in production.(my blog).

One caution for future visiters: Do not put async in `vendor.js` if you have it(the result of CommonsChunkWebpackPlugin). If both vendor.js and app.js are async, problems can happen.

choonkending:
Thanks @alubbe

PR #937, opened 2018-01-17
Add the user's cookie to Axios for pre-render fetching

tomanagle:
This is in response to #435 and #225 - where the user's session is not available to the server for pre-render data fetching. 

## Problem 
Data fetching happens before the response if sent to the user. This means that the request is coming from the server, to the server. These queries may need access to req.user, which is not available because the request has come from the server. 

## Solution
Add the user's session cookie to axios.defaults.headers.common.Cookie so passport can deserialize the user, giving access to req.user. 

## Test
Add console.log('User from request', req.user) to the all function in  /controller/topics/
Log into the application
See that the user has been attached to the request object

I also added the sessionID to the secrets file. This is because the user may change the name of this in the cookie declaration and this solution will no longer work. Adding it to a centralised location means the user can change the name of the session cookie and it will be changed everywhere that it is used.

### DIFF
--- a/config/secrets.js
+++ b/config/secrets.js
@@ -8,3 +8,5 @@ export const google = {
   callbackURL: process.env.GOOGLE_CALLBACK || '/auth/google/callback'
 };
 
+/* To make sure everything referencing the session ID knows what it is called */
+export const sessionId = 'sid';
--- a/server/init/express.js
+++ b/server/init/express.js
@@ -14,86 +14,86 @@ import { DB_TYPE, ENV } from '../../config/env';
 import { session as dbSession } from '../db';
 
 export default (app) => {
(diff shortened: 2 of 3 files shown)


slavab89:
First of all, thank you very much for the PR.
This can come instead of the PR https://github.com/reactGo/reactGo/pull/577.
Seems ok to me. Would you mind fixing the indentations though? I believe you have only 1 space of indentation set..? while our eslint (which uses airbnb is using 2 spaces

tomanagle:
Hi Slava, 

The spacing has been corrected to match eslint. 

Thank you
Tom

GGAlanSmithee:
Great work @tomanagle lgtm

PR #945, opened 2018-03-04
Upgrade packages

sebastianrjay:


sebastianrjay:
@slavab89 I've added `babel-polyfill`, as requested

slavab89:
@sebastianrjay Thanks a lot for the update! However i dont believe that the project will run.
Upgrading to webpack 4, dev-middleware etc is a breaking change that will require quite a few code changes in the config and the initialization code itself - something that is better to do in a different PR (i hope to get to it this weekend).
Moreover, you've added babel-polyfill but i dont see its usage in the project.

I suggest to try and update only patch/minor versions in the PR (with small relevant code changes) and not change major package changes, those better go into their own PR

sebastianrjay:
@slavab89 Did you try running the project? It runs for me with no issues.

If you are unable to run it, make sure the MongoDB driver is running in its own terminal tab (started via `mongod`). Then, run `rm -rf node_modules` and then `yarn && yarn dev`. (You will need to install [yarn](https://github.com/yarnpkg/yarn) if not already installed.)

> The upgrade to react 16 seems to have some requirements:
https://reactjs.org/blog/2017/09/26/react-v16.0.html#javascript-environment-requirements
which currently are not in the project (The easiest way is to include babel-polyfill)

> Moreover, you've added babel-polyfill but i dont see its usage in the project.

Do you want `babel-polyfill` in the project or not? I agree that it may not be necessary.

slavab89:
You are right, i'm sorry. I did not try to run this and it does run and seems to work
When i tried running it i got the following message:
```
WARNING in configuration
The 'mode' option has not been set. Set 'mode' option to 'development' or 'production' to enable defaults for this environment
```
On the client side i got:
```
react-dom.development.js:11524 Warning: render(): Calling ReactDOM.render() to hydrate server-rendered markup will stop working in React v17. Replace the ReactDOM.render() call with ReactDOM.hydrate() if you want React to attach to the server HTML.
```

Regarding the babel-polyfill that i've written before, something similar to (not sure about the event-source though): https://github.com/reactGo/reactGo/pull/808

PR #957, opened 2018-09-07
Remove unneeded path to bin

vikr01:
Assuming the `node_modules` is in the root of the project is not a good idea. In certain scenarios, like `yarn` `workspaces`, the `node_modules` could get hoisted up. The script will work the same without the path specification.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "repository": "https://github.com/choonkending/react-webpack-node",
   "main": "index.js",
   "scripts": {
-    "sequelize": "node_modules/.bin/sequelize",
+    "sequelize": "sequelize",
     "lint": "eslint --ext .js,.jsx app server",
     "lint:fix": "eslint --ext .js,.jsx app server --fix",
     "clean": "rimraf public && rimraf compiled",


GGAlanSmithee:
@vikr01 thanks for your contribution! But doesn't this require `sequelize` to be installed globally? I think I've had some issues with this in some situations, like when using a docker container for example. Is this safe for heroku and similar platforms?

amilajack:
By default, npm will search for `./node_modules/.bin/sequelize`. If it doesn't exist, it will fallback to global installation.

GGAlanSmithee:
@amilajack thanks for the clarification. I guess `yarn` also takes the above-mentioned `workspaces` into consideration in it's lookup algorithm, so should probably be safe to merge this.

I do have some anecdotal "evidence" that there have been problems with not explicitly specifying the path to `node_modules/bin` in my case when using a docker-based workspace (cloud9), but it could more then likely be a user error :) Will give it a few days for any additional input, then merge.

GGAlanSmithee:
thanks for your contribution @vikr01 tested locally and seems to fine.

